PR #15, opened 2024-07-16
internal: rename test

Young-Flash:
- *_test.mbt --> *_wbtest.mbt

- *_bbtest.mbt --> *_test.mbt

### DIFF
--- a/crates/moonbuild/src/fmt.rs
+++ b/crates/moonbuild/src/fmt.rs
@@ -86,7 +86,12 @@ pub fn gen_fmt(
         if pkg.is_third_party {
             continue;
         }
-        for f in pkg.files.iter().chain(pkg.wbtest_files.iter()) {
+        for f in pkg
+            .files
+            .iter()
+            .chain(pkg.wbtest_files.iter())
+            .chain(pkg.test_files.iter())
+        {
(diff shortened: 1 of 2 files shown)


crates/moonbuild/src/fmt.rs
```diff
@@ -86,7 +86,7 @@ pub fn gen_fmt(
         if pkg.is_third_party {
             continue;
         }
-        for f in pkg.files.iter().chain(pkg.test_files.iter()) {
+        for f in pkg.files.iter().chain(pkg.wbtest_files.iter()) {
```
lijunchen: also need to chain pkg.test_files?

crates/moonutil/src/module.rs
```diff
@@ -136,7 +136,7 @@ impl ModuleDB {
     pub fn validate(&self) -> anyhow::Result<()> {
         let mut errors = vec![];
         for (_, pkg) in &self.packages {
-            for item in pkg.imports.iter().chain(pkg.test_imports.iter()) {
+            for item in pkg.imports.iter().chain(pkg.wbtest_imports.iter()) {
```
lijunchen: need to validate bbtest

Issue #12, opened 2024-07-16
`moon test --target all` run all tests in parallel


bobzhang:
currently we have to run each target sequentially, we should support run targets concurrently:

```
moon test --target all
moon test --target js,wasm
```

PR #18, opened 2024-07-17
feat: support `moon test --target all`

lijunchen:
Close https://github.com/moonbitlang/moon/issues/12

### DIFF
--- a/crates/moon/src/cli/test.rs
+++ b/crates/moon/src/cli/test.rs
@@ -144,8 +144,7 @@ fn run_test_internal(
     let limit = cmd.limit;
     let sort_input = cmd.build_flags.sort_input;
 
-    let filter_package = <std::option::Option<Vec<PathBuf>> as Clone>::clone(&cmd.package)
-        .map(|it| it.into_iter().collect());
+    let filter_package = cmd.package.clone().map(|it| it.into_iter().collect());
     let filter_file = &cmd.file;
     let filter_index = cmd.index;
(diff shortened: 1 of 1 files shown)


crates/moon/tests/test_cases/mod.rs
```diff
@@ -4469,3 +4469,16 @@ fn test_import_memory_and_heap_start() {
         "#]],
     );
 }
+
+#[test]
+fn test_moon_test_target_all() {
+    let dir = TestDir::new("moon_new.in");
+    check(
+        &get_stdout_with_args_and_replace_dir(&dir, ["test", "--target", "all"]),
+        expect![[r#"
```
Young-Flash, line 4478: should add more test case with test filter and auto update, can be done in next PR

crates/moon/src/cli/test.rs
```diff
@@ -101,11 +144,12 @@ pub fn run_test(cli: UniversalFlags, cmd: TestSubcommand) -> anyhow::Result<i32>
     let limit = cmd.limit;
     let sort_input = cmd.build_flags.sort_input;
 
-    let filter_package = cmd.package.map(|it| it.into_iter().collect());
-    let filter_file = cmd.file;
+    let filter_package = <std::option::Option<Vec<PathBuf>> as Clone>::clone(&cmd.package)
```
Young-Flash:
```suggestion
    let filter_package = cmd.package.clone().map(|it| it.into_iter().collect());
```

PR #20, opened 2024-07-18
feat: support specify multiple targets

lijunchen:


crates/moon/src/cli.rs
```diff
@@ -125,8 +125,14 @@ pub struct BuildFlags {
...
 
     /// Select output target
-    #[clap(long)]
-    pub target: Option<SurfaceTarget>,
+    #[clap(long, value_delimiter = ',')]
+    pub target: Option<Vec<SurfaceTarget>>,
+
+    #[clap(skip)]
+    pub target_backend: Option<TargetBackend>,
+
```
Young-Flash, line 133:
```suggestion
Run in a serial sequence to ensure stable output results. Only valid when `--target` is also specified.
```

crates/moon/tests/test_cases/mod.rs
```diff
@@ -4486,8 +4493,8 @@ fn test_import_memory_and_heap_start() {
 }
 
 #[test]
-fn test_moon_test_target_all() {
-    let dir = TestDir::new("moon_new.in");
+fn test_many_targets() {
```
Young-Flash, line 4499: have some test case where a target backend success while other fail would be better

crates/moon/src/cli.rs
lijunchen: The CLI interface does not use this field, it is just a workaround.

PR #31, opened 2024-07-22
feat: support run single .mbt file

Young-Flash:
close https://github.com/moonbitlang/moon/issues/27

### DIFF
--- a/crates/moon/src/cli/run.rs
+++ b/crates/moon/src/cli/run.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use anyhow::{bail, Context};
 use moonbuild::dry_run;
 use moonbuild::entry;
@@ -6,6 +8,7 @@ use moonutil::common::lower_surface_targets;
 use moonutil::common::FileLock;
 use moonutil::common::RunMode;
 use moonutil::common::SurfaceTarget;
+use moonutil::common::TargetBackend;
 use moonutil::common::{MoonbuildOpt, OutputFormat};
 use moonutil::dirs::check_moon_pkg_exist;
 use moonutil::dirs::mk_arch_mode_dir;
@@ -19,8 +22,8 @@ use super::{BuildFlags, UniversalFlags};
 /// Run WebAssembly module
 #[derive(Debug, clap::Parser, Clone)]
 pub struct RunSubcommand {
-    /// The package to run
-    pub package: String,
+    /// The package or .mbt file to run
+    pub package_or_mbt_file: String,
 
     #[clap(flatten)]
     pub build_flags: BuildFlags,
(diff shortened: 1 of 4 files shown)


crates/moon/src/cli/run.rs
```diff
@@ -46,6 +49,90 @@ pub struct RunSubcommand {
...
+
+    if !mbt_file_path.exists() || !mbt_file_path.is_file() {
+        bail!("{} is not exist or not a file", mbt_file_path.display());
+    }
+
+    let file_name = mbt_file_path.file_stem().unwrap().to_str().unwrap();
+
+    let target_backend = lower_surface_targets(&cmd.build_flags.target.unwrap_or_default())
+        .first()
+        .map_or(TargetBackend::WasmGC, |it| *it);
```
lijunchen: use `TargetBackend::default()`?

crates/moon/src/cli/run.rs
```diff
@@ -46,6 +49,90 @@ pub struct RunSubcommand {
     pub args: Vec<String>,
 }
 
+fn run_single_mbt_file(cmd: RunSubcommand) -> anyhow::Result<i32> {
```
lijunchen: `moon run x.mbt --dry-run` is not working

crates/moon/src/cli/run.rs
```diff
@@ -46,6 +49,90 @@ pub struct RunSubcommand {
...
+    let target_backend = lower_surface_targets(&cmd.build_flags.target.unwrap_or_default())
+        .first()
+        .map_or(TargetBackend::WasmGC, |it| *it);
+    let core_bundle_path = moonutil::moon_dir::core_bundle(target_backend);
+
+    // `parent_path` is not always same with `parent_path`, since `cmd.package_or_mbt_file` can be something like "a/b/c/single.mbt"
+    let parent_path = mbt_file_path.parent().unwrap();
+
+    // we want all output artifacts to be in the same directory as the input single .mbt file
+    let output_core_path = &(parent_path
```
lijunchen, line 80: `moon run x.mbt` will generate `x.{mi, core, wasm}` in parent dir. Do we need to hide them?

crates/moon/src/cli/run.rs
```diff
@@ -46,6 +49,90 @@ pub struct RunSubcommand {
...
+    }
+
+    let file_name = mbt_file_path.file_stem().unwrap().to_str().unwrap();
+
+    let target_backend = lower_surface_targets(&cmd.build_flags.target.unwrap_or_default())
+        .first()
+        .map_or(TargetBackend::WasmGC, |it| *it);
+    let core_bundle_path = moonutil::moon_dir::core_bundle(target_backend);
+
+    // `parent_path` is not always same with `parent_path`, since `cmd.package_or_mbt_file` can be something like "a/b/c/single.mbt"
```
lijunchen:
```
`parent_path` is not always same with `parent_path`,
```

What does this mean?

crates/moon/src/cli/run.rs
```diff
@@ -46,6 +49,90 @@ pub struct RunSubcommand {
...
+        .wait()?;
+
+    if !moonc_link_core.success() {
+        bail!("failed to run: moonc {}", link_core_command.join(" "))
+    }
+
+    trace::scope("run", || {
+        if target_backend == TargetBackend::Wasm || target_backend == TargetBackend::WasmGC {
+            moonbuild::build::run_wat(&PathBuf::from(output_wasm_or_js_path), &cmd.args)
+        } else {
```
lijunchen: It's better to use `match` here so that if we add a new backend, we can easily find the problem.

crates/moon/tests/test_cases/run_single_mbt_file.in/a/b/c/tmp.txt
lijunchen, line 1: rename `tmp.txt` to `.gitkeep`?

crates/moon/src/cli/run.rs
lijunchen: This can be fixed by a dedicated PR.

crates/moon/src/cli/run.rs
Young-Flash: Leave it to discussion tomorrow

crates/moon/src/cli/run.rs
Young-Flash: typo, should be: `parent_path` is not always same with `current_dir`

PR #41, opened 2024-07-23
fix: report error when there are depulicate test-import alias

Young-Flash:


crates/moonbuild/src/gen/gen_check.rs
```diff
@@ -143,6 +143,20 @@ fn pkg_with_test_to_check_item(
...
+        .iter()
+        .chain(pkg.imports.iter())
+        .any(|import| {
+            import
+                .alias
+                .as_ref()
+                .map_or(false, |alias| alias.eq(pkg.last_name()))
+        })
+    {
+        bail!("Duplicate alias `{}` at \"{}\". \"test-import\" will automatically add \"import\" and current pkg as dependency so you don't need to add it manually, if you're test-importing a dependency whose default alias same as current pkg, considering give it a different alias.", pkg.last_name(), pkg.rel.components.join("/") + "/moon.pkg.json");
```
lijunchen: This line is too long and can be rewritten similarly to the following

```rust
let s = "a \
    b \
    c"
```

lijunchen: use `MOON_PKG_JSON` to replace `moon.pkg.json`

crates/moon/tests/test_cases/mod.rs
```diff
@@ -4461,6 +4461,14 @@ fn test_blackbox_failed() {
     assert!(output.contains("Package \"C\" not found in the loaded packages."));
 }
 
+#[test]
+fn test_blackbox_dedup_alias() {
+    let dir = TestDir::new("blackbox_test_dedup_alias.in");
+
+    let output = get_err_stdout_with_args_without_replace(&dir, ["test"]);
+    assert!(output.contains("error: Duplicate alias `lib` at \"lib/moon.pkg.json\". \"test-import\" will automatically add \"import\" and current pkg as dependency so you don't need to add it manually, if you're test-importing a dependency whose default alias same as current pkg, considering give it a different alias."));
```
lijunchen: Why not use `check` here?

Issue #53, opened 2024-07-23
moon check with json output


Young-Flash:
- diagnostic for moon.pkg.json & moon.mod.json
- enable ide respect alert list &  warn list config

cc @bzy-debug

PR #56, opened 2024-07-24
feat: support moon check output in json format

Young-Flash:
`moon check --output-json -q`

close https://github.com/moonbitlang/moon/issues/53

cc @bzy-debug


lijunchen (approved):
LGTM

PR #58, opened 2024-07-24
feat: add git check

Lampese:


Lampese:
Test CI seems doesn't work, is it my fault?

bzy-debug:
> Test CI seems doesn't work, is it my fault?

Might be related to secrets restriction in forked PR. I'm looking into it.

Cargo.toml
```diff
@@ -90,6 +90,7 @@ thiserror = "1.0"
 test-log = "0.2"
 fs4 = { version = "0.8.3", features = ["sync"] }
 ariadne = { version = "0.4.1", features = ["auto-color"] }
+git2 = "0.19.0"
```
lijunchen: This causes the binary size of `moon` to increase from 3.5M to 6.4M (tested on macOS), and binary size is one of our optimization goals.

Lampese: Ok, then I'll try to split a part of git2 to here

Lampese:
Test CI seems doesn't work again? @bzy-debug

crates/moonutil/src/git.rs
```diff
@@ -0,0 +1,50 @@
...
+        .stderr(std::process::Stdio::null())
+        .current_dir(path)
+        .status();
+    match output {
+        Ok(out) => out.success(),
+        _ => false,
+    }
+}
+
+pub fn create_git_repo(path: &Path) {
```
Young-Flash: a better name would be `git_init_repo`

crates/moon/tests/test_cases/mod.rs
```diff
@@ -489,7 +489,6 @@ fn test_moon_new_exist() {
     );
 
     assert!(res.contains("Created hello"));
-    assert!(res.contains("Initialized empty Git repository"));
```
Young-Flash: IMO this could be retained to verify that `git init` have been executeed

Young-Flash (approved):
Mostly LGTM, thanks

Lampese:
IMO this coverage check action may be some wrong?

PR #60, opened 2024-07-24
internal: tweak moon run single .mbt file

Young-Flash:


crates/moon/src/cli/run.rs
```diff
@@ -202,7 +206,18 @@ pub fn run_run_internal(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Res
 
     let sort_input = cmd.build_flags.sort_input;
 
-    let package_path = cmd.package_or_mbt_file.clone();
+    // run .mbt inside a package should run as a package
+    let package_path = if cmd.package_or_mbt_file.ends_with(".mbt") {
+        // `package_path` based on `source_dir`
+        dunce::canonicalize(std::env::current_dir()?.join(cmd.package_or_mbt_file))?
```
lijunchen: canonicalize a nonexistent path will cause an error, add context for better error reporting

crates/moon/src/cli/run.rs
```diff
@@ -90,15 +91,14 @@ fn run_single_mbt_file(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Resu
         .map_or(TargetBackend::default(), |it| *it);
     let core_bundle_path = moonutil::moon_dir::core_bundle(target_backend);
 
-    // `parent_path` is not always same with `current_dir`, since `cmd.package_or_mbt_file` can be something like "a/b/c/single.mbt"
-    let parent_path = mbt_file_path.parent().unwrap();
+    let output_artifact_path = std::env::temp_dir().join("moon_run_artifact");
```
lijunchen: use `tempfile` crate?

crates/moon/src/cli/run.rs
```diff
@@ -76,7 +77,7 @@ pub fn run_run(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Result<i32>
 }
 
 fn run_single_mbt_file(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Result<i32> {
-    let current_dir = std::env::current_dir().unwrap();
+    let current_dir = std::env::current_dir()?;
     let mbt_file_path = current_dir.join(cmd.package_or_mbt_file);
 
     if !mbt_file_path.exists() || !mbt_file_path.is_file() {
```
lijunchen: `if !mbt_file_path.exists() || !mbt_file_path.is_file()`

=> 

`if !mbt_file_path.is_file()`

crates/moon/src/cli/run.rs
```diff
@@ -117,7 +117,7 @@ fn run_single_mbt_file(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Resu
     let link_core_command = [
         "link-core",
         &core_bundle_path.join("core.core").display().to_string(),
```
lijunchen: `core_bundle_path.join("core.core")` => `core_core()`

crates/moon/src/cli/run.rs
lijunchen: and no need to create `moon_run_artifact` if use `tempfile`

Young-Flash: we want the artifact to be placed in a known path, ide debug need it

lijunchen: If so, I suggest placing temp files in `.moon`.

lijunchen: It seems that `std::env::temp_dir()` returns the same directory each time. What will happen if there is a name conflict in `moon_run_artifact`?

PR #67, opened 2024-07-25
FEAT: Add Shell Completion

notch1p:
adds go-like shell completion for bash/elvish/fish/pwsh/zsh.
### example
```bash
moon shell-completion --shell={SHELL}
```
Generates shell completion script at runtime. By default it reads from `$SHELL` to determine the type of shell.

image

record

### Usage

```bash
moon shell-completion --shell={SHELL}
```

The `shell-completion` subcommand generates shell completion scripts for bash/elvish/fish/powershell/zsh.

Use `moon shell-cimpletion` to output a completion script. By default if `--shell` isn't provided explicitly, `moon` will read from `$SHELL` to determine the correct shell type. Providing any of `{bash,elvish,fish,powershell,zsh}` to `--shell` would override this.

Here's how one may use it w/ Zsh:

Zsh autoloads function from `$FPATH`, we could add `$ZSH/completions` to it, and using
```bash
moon shell-completion --shell=zsh > $ZSH/completions/_moon
```
will enable `moon` completions for Zsh.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ thiserror = "1.0"
 test-log = "0.2"
 fs4 = { version = "0.8.3", features = ["sync"] }
 ariadne = { version = "0.4.1", features = ["auto-color"] }
+clap_complete = { version = "4.5.4" }
 
 [profile.release]
 debug = false
--- a/crates/moon/src/cli/shell_completion.rs
+++ b/crates/moon/src/cli/shell_completion.rs
@@ -1,0 +1,40 @@
+// moon: The build system and package manager for MoonBit.
+// Copyright (C) 2024 International Digital Economy Academy
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+// For inquiries, you can contact us via e-mail at jichuruanjian@idea.edu.cn.
+
+use super::MoonBuildCli;
+use clap::CommandFactory;
+use clap_complete::{generate, Shell};
+use moonutil::cli::UniversalFlags;
+use std::io;
+
+/// Generate shell completion for bash/elvish/fish/pwsh/zsh to stdout
+#[derive(Debug, clap::Parser)]
+pub struct ShellCompSubCommand {
+    /// The shell to generate completion for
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -89,6 +89,7 @@ thiserror = "1.0"
 test-log = "0.2"
 fs4 = { version = "0.8.3", features = ["sync"] }
 ariadne = { version = "0.4.1", features = ["auto-color"] }
+clap_complete = { version = "4.5.4" }
```
Young-Flash, line 92: how much will this increase the moon binary size?

notch1p: As of version `0.1.20240725 (5e7a3e8 2024-07-25)` 

- original `moon` is 3,691,720B (~3.7M)
- w/ shell completion is 3,741,640B (~3.7M) which is roughly 1% larger

I would say pretty acceptable

crates/moon/src/cli/shell_completion.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Generate shell completion for bash/elvish/fish/pwsh/zsh to stdout
+#[derive(Debug, clap::Parser)]
+pub struct ShellCompSubCommand {
+    /// The shell to generate completion for
+    #[clap(value_enum, long, ignore_case = true, value_parser = clap::builder::EnumValueParser::<Shell>::new(), default_value_t = Shell::from_env().unwrap_or(Shell::Bash), value_name = "SHELL")]
+    pub shell: Shell,
+}
+
+pub fn gen_shellcomp(_cli: &UniversalFlags, cmd: ShellCompSubCommand) -> anyhow::Result<i32> {
+    if _cli.dry_run {
```
lijunchen, line 34: why add `_` before `cli` and `moon`?

lijunchen (commented):
1. 
```
moon build --source
```

will auto-complete to

```
moon build --source-dir=
```

Currently `=` does not work for `--source-dir`

2. Please add documentation for Unix/Windows users in the current PR description so we can easily copy it into the weekly release notes.

3. Do not introduce merge commits. For more details, see https://github.com/moonbitlang/moon/blob/main/docs/dev/README.md#maintain-semi-linear-history

crates/moon/src/cli/shell_completion.rs
notch1p: `_cli` is copied from `new.rs`; `_moon` really has no reason, I didn't really think that much, if needed I could change back to `cli` and `moon`.

notch1p:
Can you elaborate on problem 1? technically `--source-dir <dir>` and `--source-dir=<dir>` works the same, at least on my machine

lijunchen:
> Can you elaborate on problem 1? technically `--source-dir <dir>` and `--source-dir=<dir>` works the same, at least on my machine, I just tested it, even with the release version of `moon`

On macOS

```
$ cd ~
$ moon new hello

$ moon build --source-dir hello
Finished. moon: ran 3 tasks, now up to date

$ moon build --source-dir ~/hello
Finished. moon: no work to do

$ moon build --source-dir=hello
Finished. moon: no work to do

$ moon build --source-dir=~/hello
error: failed to set source directory

Caused by:
    No such file or directory (os error 2)
```

notch1p:
This is mostly a shell problem. `=~` is a bash operator for matching strings. In this case one has to use
```
moon build --source-dir=${HOME}/hello
```

lijunchen (approved):
LGTM. Please use squash merge, since your branch contains a meaningless commit.

image

PR #74, opened 2024-07-26
chore: better upgrade output

qazxcdswe123:
## Type of Pull Request

- [x] Other (please describe): output formatting

- [x] No

- [x] No

## Checklist:

### DIFF
--- a/crates/moonbuild/src/upgrade.rs
+++ b/crates/moonbuild/src/upgrade.rs
@@ -289,10 +289,9 @@ pub fn do_upgrade(root: &'static str) -> Result<i32> {
 
                             // use new moon to bundle
                             let moon = moon_dir::home().join("bin").join("moon");
-                            println!("moon version:");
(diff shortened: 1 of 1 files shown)


lijunchen (approved):

crates/moonbuild/src/upgrade.rs
```diff
@@ -289,10 +289,9 @@ pub fn do_upgrade(root: &'static str) -> Result<i32> {
 
                             // use new moon to bundle
                             let moon = moon_dir::home().join("bin").join("moon");
-                            println!("moon version:");
                             println!("Compiling {} ...", MOONBITLANG_CORE);
                             let out = std::process::Command::new(&moon).args(["version"]).output()?;
```
Young-Flash, line 293: we could have `moon version --all` here to show more info, and the following could be "moonbit toolchain version"

Issue #47, opened 2024-07-23
test block level parallelism


lijunchen:

Young-Flash:
we could first make package level parallelism

cc @bobzhang

PR #98, opened 2024-08-01
feat: parallelize moon test in pkg level

Young-Flash:
test at [core](https://github.com/moonbitlang/core) with the following result

before: 2.2 seconds on average (after warmup)
image

after: 0.2 seconds on average (after warmup)
image

part of https://github.com/moonbitlang/moon/issues/47

### DIFF
--- a/crates/moon/src/cli/test.rs
+++ b/crates/moon/src/cli/test.rs
@@ -78,6 +78,9 @@ pub struct TestSubcommand {
 
     #[clap(flatten)]
     pub auto_sync_flags: AutoSyncFlags,
+
+    #[clap(long)]
+    pub no_parallelize: bool,
 }
 
 pub fn run_test(cli: UniversalFlags, cmd: TestSubcommand) -> anyhow::Result<i32> {
@@ -184,6 +187,7 @@ fn run_test_internal(
         run_mode,
         quiet: true,
         verbose: cli.verbose,
+        no_parallelize: cmd.no_parallelize,
         ..Default::default()
     };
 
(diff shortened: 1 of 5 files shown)


Young-Flash:
test failure supported to be fixed in the next release

crates/moon/src/cli/test.rs
```diff
@@ -78,6 +78,9 @@ pub struct TestSubcommand {
 
     #[clap(flatten)]
     pub auto_sync_flags: AutoSyncFlags,
+
+    #[clap(long)]
+    pub no_parallelize: bool,
```
lijunchen, line 83: just use `no-parallel`

Maybe we need `--jobs` flag?

crates/moonbuild/src/entry.rs
```diff
@@ -321,101 +328,132 @@ pub fn run_test(
...
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()?;
+
+    let mut handlers = vec![];
 
     for runnable_artifact in runnable_artifacts.iter() {
         let p = Path::new(runnable_artifact);
 
+        let passed = passed.clone();
```
lijunchen, line 346: use `Arc::clone` instead of `.clone`?

lijunchen (approved):

crates/moonbuild/src/entry.rs
Young-Flash: seems no difference here, ide inlay hints will tell us it's `Arc` wrapped

crates/moon/src/cli/test.rs
Young-Flash: will make at next PR

PR #102, opened 2024-08-01
warn clippy::clone_on_ref_ptr

lijunchen:
https://rust-lang.github.io/rust-clippy/master/#/clone_on_ref_ptr

cc @Young-Flash

### DIFF
--- a/crates/moonutil/src/lib.rs
+++ b/crates/moonutil/src/lib.rs
@@ -16,6 +16,8 @@
 //
 // For inquiries, you can contact us via e-mail at jichuruanjian@idea.edu.cn.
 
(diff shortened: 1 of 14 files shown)


crates/moonutil/src/lib.rs
```diff
@@ -16,6 +16,8 @@
 //
 // For inquiries, you can contact us via e-mail at jichuruanjian@idea.edu.cn.
 
+#![warn(clippy::clone_on_ref_ptr)]
```
Young-Flash, line 19: can we make this a workspace level config?

lijunchen: have no idea how to do this🤣

PR #128, opened 2024-08-06
feat: support module level alert&warn list config

Young-Flash:
will add a test when `moonc` fix the 'warn list' overwrite problem


crates/moonutil/src/scan.rs
```diff
@@ -274,6 +274,18 @@ fn scan_one_package(
         wbtest_mbt_files.sort();
         test_mbt_files.sort();
     }
+    // if warn_list & alert_list not set in current moon.pkg.json, use the one in moon.mod.json
+    let warn_list = if pkg.warn_list.is_none() {
```
lijunchen: Should pkg.warn_list be merged with mod's warn_list or just replaced?

Young-Flash: let's append

Issue #170, opened 2024-08-14
Compile errors and warnings are printed to stdout; should be stderr


lynzrand:
# Bug Report

### OS

Operating System: Linux

### MoonBit CLI Tools Version

```
moon 0.1.20240812 (e15eb21 2024-08-12) ~/.moon/bin/moon
moonc v0.1.20240812+356f388af ~/.moon/bin/moonc
moonrun 0.1.20240716 (08bce9c 2024-07-16) ~/.moon/bin/moonrun
```

## Steps to Reproduce

`moon run <program>` with program having some compile warnings.

## Expected Behavior

The program runs normally, and warnings are printed to stderr.

## Actual Behavior

Warnings are printed to stdout, interfering with the actual program output.

PR #171, opened 2024-08-14
Move diagnostic printing to stderr

lynzrand:
## Related Issues

- [ ] Related issues: #170 

## Type of Pull Request

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

  - Diagnostics will be printed to stderr instead of stdout 
  
    > In most cases, [Report::eprint](https://docs.rs/ariadne/latest/ariadne/struct.Report.html#method.eprint) is the [‘more correct’](https://en.wikipedia.org/wiki/Standard_streams#Standard_error_(stderr)) function to use. 
    >
    > - [`ariadne::Report::print`](https://docs.rs/ariadne/latest/ariadne/struct.Report.html#method.print)
- [ ] No

- [x] No

## Checklist:

### DIFF
--- a/crates/moon/tests/test_cases/mod.rs
+++ b/crates/moon/tests/test_cases/mod.rs
@@ -1544,13 +1544,6 @@ fn test_moon_test_succ() {
             ["test", "-v", "--sort-input", "--no-parallelize"],
         ),
         expect![[r#"
-            Warning: [1001]
-               ╭─[$ROOT/lib2/nested/lib.mbt:1:4]
-               │
-             1 │ fn add1(x : Int) -> Int {
-               │    ──┬─  
-               │      ╰─── Warning: Unused function 'add1'
-            ───╯
             test moontest/lib/hello_wbtest.mbt::0 ok
             test moontest/lib2/hello_wbtest.mbt::0 ok
             test moontest/lib2/nested/lib_wbtest.mbt::0 ok
@@ -3259,7 +3252,10 @@ fn test_warn_list_real_run() {
     let dir = TestDir::new("warn_list.in");
 
     check(
-        &get_stdout_with_args_and_replace_dir(&dir, ["build", "--sort-input", "--no-render"]),
+        &get_stderr_on_success_with_args_and_replace_dir(
+            &dir,
+            ["build", "--sort-input", "--no-render"],
+        ),
         expect![[r#"
             Finished. moon: ran 4 tasks, now up to date
         "#]],
@@ -3273,7 +3269,10 @@ fn test_warn_list_real_run() {
     );
 
     check(
-        &get_stdout_with_args_and_replace_dir(&dir, ["bundle", "--sort-input", "--no-render"]),
+        &get_stderr_on_success_with_args_and_replace_dir(
+            &dir,
+            ["bundle", "--sort-input", "--no-render"],
+        ),
         expect![[r#"
             Finished. moon: ran 4 tasks, now up to date
         "#]],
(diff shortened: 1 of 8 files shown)


crates/moon/tests/test_cases/mod.rs
```diff
@@ -5282,11 +5281,17 @@ fn test_moon_check_json_output() {
         "#]],
         );
         check(
-            &get_stdout_with_args_and_replace_dir(&dir, ["check", "--output-json"]),
```
lynzrand, line 5281: Should this be on stdout or stderr?

It won't interfere with program running, so either is okay for this case (nobody will output JSON when doing `moon run` right?). However all other compile results reside in stderr, so moving this output to stderr for consistency _is possible_.

Young-Flash: cc @bzy-debug

Young-Flash:
need rebase and resolve conflicts

crates/moon/tests/test_cases/mod.rs
```diff
@@ -5296,11 +5304,17 @@ fn test_moon_check_json_output() {
...
+            &get_stderr_on_success_with_args_and_replace_dir(
+                &dir,
+                ["check", "--output-json", "-q"],
+            ),
+            expect![""],
+        );
+        check(
+            &get_stderr_on_success_with_args_and_replace_dir(&dir, ["check", "--output-json"]),
             expect![[r#"
-            {"$message_type":"diagnostic","level":"warning","loc":{"path":"$ROOT/main/main.mbt","start":{"line":3,"col":3,"offset":27},"end":{"line":3,"col":10,"offset":34}},"message":"Warning (Alert alert_2): alert_2","error_code":2000}
```
Young-Flash, line 5301: should keep

crates/moon/tests/test_cases/specify_source_dir_001.in/moon.test
```diff
@@ -68,6 +69,7 @@
     "source": "src"
   }
   $ moon build
+
```
Young-Flash, line 72: why add the newline 😯

crates/moon/tests/test_cases/mod.rs
lynzrand: I suppose this should be in stdout in this case (instead of also migrated to stderr), in order to keep CLI, language server, etc. stable.

crates/moon/tests/test_cases/mod.rs
```diff
@@ -3858,7 +3860,7 @@ fn test_multi_process() {
 
                 if output.status.success() {
                     success.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
-                    let out = String::from_utf8(output.stdout).unwrap();
+                    let out = String::from_utf8(output.stderr).unwrap();
```
Young-Flash, line 3863: should this change to stderr?

lynzrand: Yes, see below: the "no work to do/now up to date" line now resides in stderr instead of stdout (to not interfere with the actual run output), so it needs be changed here.

crates/moon/tests/test_cases/specify_source_dir_001.in/moon.test
lynzrand: See xtask/src/cmdtest/exec.rs:76 : If stdout is empty and stderr is not, the result will be `"\n{stderr}"`.

```rust
impl ExecResult {
    pub fn normalize(&self, workdir: &Path) -> String {
        let actual = if self.stderr.is_empty() {
            self.stdout.clone()
        } else {
            format!("{}\n{}", self.stdout, self.stderr) // <-
        };
        let actual = if self.exit_code != 0 {
            format!("[{}]\n{}", self.exit_code, actual)
        } else {
            actual
        };
        let actual = replace_dir(&actual, &workdir);
        actual
    }
}
```

lijunchen (approved):

PR #212, opened 2024-08-26
internal: support test block debug with sourcemap

Young-Flash:
with `--debug` flag, `--source-map` will be pass to `build-package` & `link-core`

cc @hackwaly


hackwaly:
I've tested locally. `.map` files generated as expected.

PR #213, opened 2024-08-26
internal: move moon.db into corrsponding run mode dir

Young-Flash:


crates/moonutil/src/dirs.rs
```diff
@@ -233,6 +234,7 @@ pub fn clean_dir_in_target(target_dir: &Path) -> anyhow::Result<()> {
 }
 
 pub fn has_moon_db(target_dir: &Path) -> bool {
+    // println!("target_dir: {:?}", target_dir);
```
lijunchen: delete

crates/moonutil/src/dirs.rs
```diff
@@ -129,12 +134,8 @@ pub fn mk_arch_mode_dir(
         recreate_moon_db(source_dir, &arch_mode_dir)?;
         clean_dir_in_target(&arch_mode_dir)?;
     }
-
-    let arch_mode_dir = arch_mode_dir.join(mode.to_dir_name());
-
-    if !arch_mode_dir.exists() {
-        std::fs::create_dir_all(&arch_mode_dir).expect("Failed to create target directory");
-    }
+    // seems RAII drop will not always unlock the file lock correctly so manually unlock here
```
lijunchen: Why not implement drop for FileLock?

PR #234, opened 2024-08-29
fix generate test driver

Young-Flash:
generate test driver will make driver for a package, we shouldn only pass `--package` to that subcommand, `--file` & `--index` is unnecessary and cause regenerate.

### DIFF
--- a/crates/moon/tests/test_cases/mod.rs
+++ b/crates/moon/tests/test_cases/mod.rs
@@ -4485,16 +4485,16 @@ fn test_blackbox_success() {
         ),
         expect![[r#"
             moonc build-package ./D/hello.mbt -o ./target/wasm-gc/debug/test/D/D.core -pkg username/hello/D -pkg-sources username/hello/D:./D -target wasm-gc -g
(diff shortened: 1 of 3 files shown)


crates/moon/tests/test_cases/mod.rs
```diff
@@ -5972,3 +5972,45 @@ fn test_js() {
...
+            "moonbitlang/hello/lib",
+            "--file",
+            "hello_wbtest.mbt",
+            "--index",
+            "0",
+        ],
+    );
+    let time_3 = driver_file.metadata().unwrap().modified().unwrap();
+
+    assert!(time_2 == time_3);
```
lijunchen, line 6066: add a counterexample?
for example, add a new mbt file to force rebuilding, then we should have something like `time_4 != time_3`

crates/moonbuild/src/gen/gen_runtest.rs
```diff
@@ -963,7 +955,7 @@ fn gen_generate_test_driver_command(
     .arg(&moonbuild_opt.source_dir.display().to_string())
     .arg("--target-dir")
     .arg(&moonbuild_opt.target_dir.display().to_string())
-    .args(test_filter_command)
+    .args(["--package", &item.package_name])
```
lijunchen, line 958: also need to clean up file and index field in `struct GeneratedTestDriverSubcommand`

PR #242, opened 2024-08-30
feat: support build graph visualization

Young-Flash:
image

close https://github.com/moonbitlang/moon/issues/229


crates/moon/tests/test_cases/mod.rs
```diff
@@ -437,6 +437,7 @@ fn test_moon_help() {
               -v, --verbose                  Increase verbosity
                   --trace                    Trace the execution of the program
                   --dry-run                  Do not actually run the command
+                  --build-graph              generate build graph
```
lijunchen, line 440: need to update help message in /README.md and manual, manual-zh

crates/moon/tests/test_cases/mod.rs
```diff
@@ -3556,11 +3557,12 @@ fn test_deny_warn() {
...
 
-    let s = std::str::from_utf8(&out).unwrap().to_string();
-
-    assert!(s.contains(
-        "failed: moonc check -error-format json -w @a -alert @all-raise-throw-unsafe+deprecated"
-    ));
+    check(
+        std::str::from_utf8(&out).unwrap(),
+        expect![[r#"
+        failed: check: username/hello/lib
```
lijunchen: Can we print the command when it fails? We can quickly check whether it is a moon issue or a moonc issue.

crates/moon/tests/test_cases/moon_build_package.in/moon.test
```diff
@@ -2,11 +2,11 @@
...
-  $ moon build
+  $ moon build --build-graph
   
   Finished. moon: ran 2 tasks, now up to date
   
   $ xls ./target/wasm-gc/release/build/lib/
   lib.core lib.mi
   $ xls ./target/wasm-gc/release/build/
-  .moon-lock build.moon_db build.output hello.core hello.mi lib moon.db
+  .moon-lock build.moon_db build.output build_graph.dot hello.core hello.mi lib moon.db
```
lijunchen, line 13: use xcat to snapshot dot file?

```
  $ xcat ./target/wasm-gc/release/build/build_graph.dot
```

crates/moon/src/cli/build.rs
```diff
@@ -130,6 +130,7 @@ fn run_build_internal(
         run_mode,
         quiet: cli.quiet,
         verbose: cli.verbose,
+        build_graph: cli.build_graph,
```
lijunchen, line 133: disable `--build-graph` while `--dry-run`?

crates/moon/tests/test_cases/mod.rs
Young-Flash: need to modify n2
https://github.com/moonbitlang/n2/blob/d5bc0ccafbfb4b76eab7d1036746ad93bff59e45/src/progress.rs#L18-L24

lijunchen: It would be nice to keep the original behavior.

Or, can we add description fields only when passing `--build-graph`?

crates/moonbuild/src/entry.rs
```diff
@@ -107,6 +110,10 @@ pub fn n2_run_interface(
             });
     };
 
+    if moonbuild_opt.build_graph {
+        vis_build_graph(&state, moonbuild_opt);
```
lijunchen, line 114: Do we need to print messages to tell the user the location of the outputted dot file?

lijunchen (approved):

PR #273, opened 2024-09-09
internal: unify test driver template

Young-Flash:
export `execute` in wasm, unify the test driver for wasm | wasmgc and js backend, put the difference into `wasm_args.mbt` and `js_args.mbt`.

this PR modify `moonrun` so ci is expected to be failed


Young-Flash:
should be merged after today's release

PR #276, opened 2024-09-09
Trace correct test time

lgxbslgx:
Currently, when using the command `moon test --trace` to record the test time, the output in `trace.json` is not right. The duration is always `0` (shown below).

```
,{"pid":0, "name":"test", "ts":7753866, "tid": 0, "ph":"X", "dur":0}
,{"pid":0, "name":"test", "ts":7758979, "tid": 0, "ph":"X", "dur":0}
```

It is because the code in [entry.rs](https://github.com/moonbitlang/moon/blob/1caf7785baf1f436f21697f1d5993fb53aa6347e/crates/moonbuild/src/entry.rs#L559..L569) just records the time to construct a `Future` instead of the time to run the test.

This PR relies on https://github.com/moonbitlang/n2/pull/2 and can't be integrated before it. After merging these two PRs, the output should be right (as shown below).

```
,{"pid":0, "name":"test", "ts":9205546, "tid": 0, "ph":"X", "dur":401710}
,{"pid":0, "name":"test", "ts":9196616, "tid": 0, "ph":"X", "dur":414907}
```

## Related Issues

- [ ] Related issues: #____

## Type of Pull Request

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

  - record the right test time
- [ ] No

- [x] No

## Checklist:

### DIFF
--- a/crates/moonbuild/src/entry.rs
+++ b/crates/moonbuild/src/entry.rs
@@ -551,16 +551,16 @@ pub fn run_test(
 
             let printed = Arc::clone(&printed);
             handlers.push(async move {
-                let mut result = trace::scope("test", || async {
+                let mut result = trace::async_scope(
+                    "test",
                     execute_test(
                         moonc_opt.build_opt.target_backend,
                         &artifact_path,
                         target_dir,
                         &test_args,
                         &file_test_info_map,
-                    )
-                    .await
-                })
+                    ),
+                )
                 .await;
                 match result {
                     Ok(ref mut test_res_for_cur_pkg) => {


lgxbslgx:
What about this PR? I have another draft patch locally which is related to this patch and has git conflict with this patch. Kindly ping for review. Thanks a lot.

crates/moonbuild/src/entry.rs
```diff
@@ -551,16 +551,16 @@ pub fn run_test(
 
             let printed = Arc::clone(&printed);
             handlers.push(async move {
-                let mut result = trace::scope("test", || async {
+                let mut result = trace::async_scope(
+                    "test",
```
Young-Flash, line 555: we could sepecify which test is running

Young-Flash (approved):

crates/moonbuild/src/entry.rs
lgxbslgx: Several names which are passed to `trace::scope` need to be adjusted/polished. It is good to fix them in another patch.

PR #289, opened 2024-09-12
add test hints

lijunchen:
moon test is different from other modes, this PR only adds hints for test

### DIFF
--- a/crates/moon/src/cli/test.rs
+++ b/crates/moon/src/cli/test.rs
@@ -91,13 +91,14 @@ pub fn run_test(cli: UniversalFlags, cmd: TestSubcommand) -> anyhow::Result<i32>
     } = cli.source_tgt_dir.try_into_package_dirs()?;
 
     if cmd.build_flags.target.is_none() {
-        return run_test_internal(&cli, &cmd, &source_dir, &target_dir);
+        return run_test_internal(&cli, &cmd, &source_dir, &target_dir, None);
     }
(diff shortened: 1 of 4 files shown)


crates/moon/src/cli/test.rs
```diff
@@ -91,13 +91,26 @@ pub fn run_test(cli: UniversalFlags, cmd: TestSubcommand) -> anyhow::Result<i32>
...
     if cmd.build_flags.target.is_none() {
-        return run_test_internal(&cli, &cmd, &source_dir, &target_dir);
+        return run_test_internal(&cli, &cmd, &source_dir, &target_dir, None);
     }
     let surface_targets = cmd.build_flags.target.clone().unwrap();
     let targets = lower_surface_targets(&surface_targets);
     if cmd.update && targets.len() > 1 {
         return Err(anyhow::anyhow!("cannot update test on multiple targets"));
     }
+    let display_backend_hint: Option<u16> = if targets.len() > 1 {
```
Young-Flash: could we remove if we put backend into end ?
```
      Total tests: 0, passed: 0, failed: 0. [wasm]
      Total tests: 0, passed: 0, failed: 0. [wasm-gc]
      Total tests: 0, passed: 0, failed: 0. [js]
```

lijunchen: In the current situation, this is a good idea.

PR #290, opened 2024-09-12
internal: add compile conds into packages.json for ide

Young-Flash:


Young-Flash:
merged this first, to integrate with the IDE

PR #296, opened 2024-09-12
Improve parallelism for 'moon test'

lgxbslgx:
Currently, when using the command `moon test`, `moon` will create 16 threads by using the `tokio` library. But these 16 threads will never be used after creating. And these 16 threads become more strange when combining with the argument `--no-parallel`. It is a misuse of the library `tokio`.

This patch, which want to fix the issue, contains the following features:
- When the `--no-parallel` is not provided, run the tests parallelly
  - the thread count is equal to the logic cores (not the constant 16)
  - spawn the test tasks to the `tokio` runtime scheduler (use the threads `tokio` created to run the test)
- When the `--no-parallel` is provided,  run the tests sequentially
  - don't create new thread
  - run the test one by one in current thread

Thanks for your review.

## Related Issues

- [ ] Related issues: #____

## Type of Pull Request

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] Other (please describe): Improve performance

- [ ] No

- [x] No

## Checklist:

### DIFF
--- a/crates/moonbuild/src/entry.rs
+++ b/crates/moonbuild/src/entry.rs
@@ -448,24 +448,20 @@ fn convert_moonc_test_info(
 
 #[allow(clippy::too_many_arguments)]
 pub fn run_test(
-    moonc_opt: &MooncOpt,
-    moonbuild_opt: &MoonbuildOpt,
+    moonc_opt: MooncOpt,
+    moonbuild_opt: MoonbuildOpt,
     build_only: bool,
     test_verbose_output: bool,
     auto_update: bool,
-    module: &ModuleDB,
+    module: ModuleDB,
 ) -> anyhow::Result<Vec<Result<TestStatistics, TestFailedStatus>>> {
-    let target_dir = &moonbuild_opt.target_dir;
-    let state = crate::runtest::load_moon_proj(module, moonc_opt, moonbuild_opt)?;
+    let moonc_opt = Arc::new(moonc_opt);
+    let moonbuild_opt = Arc::new(moonbuild_opt);
+    let module = Arc::new(module);
 
-    let result = n2_run_interface(state, moonbuild_opt)?;
+    let state = crate::runtest::load_moon_proj(&module, &moonc_opt, &moonbuild_opt)?;
+    let result = n2_run_interface(state, &moonbuild_opt)?;
     render_result(result, moonbuild_opt.quiet, "testing")?;
-
-    let runtime = tokio::runtime::Builder::new_multi_thread()
-        // todo: add config item
-        .worker_threads(16)
-        .enable_all()
-        .build()?;
 
     let mut handlers = vec![];
 
@@ -490,7 +486,10 @@ pub fn run_test(
         }
 
         // convert moonc test info
-        let test_info_file = target_dir.join(pkg.rel.fs_full_name()).join(TEST_INFO_FILE);
+        let test_info_file = moonbuild_opt
+            .target_dir
+            .join(pkg.rel.fs_full_name())
+            .join(TEST_INFO_FILE);
         let current_pkg_test_info = convert_moonc_test_info(
             &test_info_file,
             pkg,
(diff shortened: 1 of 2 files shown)


lgxbslgx:
I compare the run time of the main branch and this patch in the standard library [core](https://github.com/moonbitlang/core)  by using the command `time moon test` simply. Unfortunately, their run time are almost the same. My local env is 2 cores & 4 theads. Could you help me test the performance in a machine with multi cores (if you have)? Thanks.

crates/moonbuild/src/entry.rs
```diff
@@ -448,25 +448,21 @@ fn convert_moonc_test_info(
 
 #[allow(clippy::too_many_arguments)]
 pub fn run_test(
-    moonc_opt: &MooncOpt,
-    moonbuild_opt: &MoonbuildOpt,
+    moonc_opt: MooncOpt,
+    moonbuild_opt: MoonbuildOpt,
```
Young-Flash, line 452: why change the param type from &MoonbuildOpt to MoonbuildOpt and use Arc::clone() below?

crates/moonbuild/src/entry.rs
```diff
@@ -600,15 +602,41 @@ pub fn run_test(
...
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(thread_count)
+            .enable_all()
+            .build()?;
+
+        runtime.block_on(async {
+            let mut res_handlers = vec![];
+            for handler in handlers {
+                // Submit tasks to the scheduler
+                res_handlers.push(runtime.spawn(handler));
```
Young-Flash, line 626: do we need another `res_handlers` and `spawn` here? IMO the previous `runtime.block_on(futures::future::join_all(handlers))` is easy to understand

crates/moonbuild/src/entry.rs
```diff
@@ -600,15 +602,41 @@ pub fn run_test(
...
+                // Tasks are run sequentially by using the `await` expression directly.
                 results.push(handler.await);
             }
             results
         })
     } else {
-        runtime.block_on(futures::future::join_all(handlers))
+        let thread_count = match std::thread::available_parallelism() {
+            Ok(n) => n.get(),
+            Err(_) => num_cpus::get(),
```
Young-Flash: could we just set `thread_count` to 1 in here, so we can remove `num_cpus` dep

cc @lijunchen

crates/moonbuild/src/entry.rs
lgxbslgx: When using `runtime.block_on(futures::future::join_all(handlers))`, the threads created by `tokio` are never used. Because the tasks are never spawned/submitted to the scheduler, all of them are run in the current thread. You can debug the code or/and print the thread id in method `entry.rs::execute_test` to verify it.

crates/moonbuild/src/entry.rs
lgxbslgx: `MoonbuildOpt`(and `MooncOpt` and `ModuleDB`) need to be shared in multiple threads, so `Arc` is needed. And the `Arc::new` want to move the value instead of getting a reference, so I change the parameter.

crates/moonbuild/src/entry.rs
lgxbslgx: I notice the default worker thread count in `tokio` is the number of CPUs and it also uses the method `std::thread::available_parallelism`. If `std::thread::available_parallelism` fails, the thread count will default to `1`. The related invocation link is shown below.
```
tokio::runtime::builder::Builder::build ->
tokio::runtime::builder::Builder::build_threaded_runtime ->
tokio::loom::sys::num_cpus
```

> could we just set thread_count to 1 in here, so we can remove num_cpus dep

So I agree with your idea and propose not to set the worker thread count munually (just use the default value in `tokio`).

Young-Flash (approved):
Thanks!

PR #346, opened 2024-09-26
fix update expect test

Young-Flash:
- when updating expect test, if rerun test success, we should update the test resut directly and `continue` to process the next test result in the `for` loop
- `TestArgs` should be passing to `js_driver.js` instead of hardcoding into it


lijunchen (approved):

Young-Flash:
we should keep the ability to run test artifact without cli args for js backend, which used in ide debug
https://github.com/moonbitlang/moon/pull/351

PR #351, opened 2024-09-29
fix js test args for debug

Young-Flash:
this PR enable running test artifact without cli args for js backend, so ide could use `node` to run `xx.cjs` directly

cc @hackwaly


hackwaly:
lgtm.

PR #359, opened 2024-10-07
chore: remove invalid Instant

Lampese:
I would like to remove Instant from the glue code for the following reasons:

- Instant is no binding on the Rust side, instead __moonbit_time_unstable has.

- From a consistency perspective, __moonbit_time_unstable is the one that should be retained.

### DIFF
--- a/crates/moonrun/src/template/js_glue.js
+++ b/crates/moonrun/src/template/js_glue.js
@@ -8,14 +8,6 @@ const spectest = {
         print_char: (x) => print(x),
         read_char: () => read_char(),
     },
-    Instant: {
-        now() {
-            return instant_now();
-        },
-        elapsed_as_secs_f64(instant) {
-            return instant_elapsed_as_secs_f64(instant);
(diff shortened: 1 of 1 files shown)


Young-Flash:
the ci problem maybe related to bleeding moonc

tonyfettes:
> maybe @tonyfettes need instant_now?

No.

Lampese:
> - the ci problem maybe related to bleeding moonc
> - maybe @tonyfettes need `instant_now`?
> https://github.com/moonbitlang/moon/blob/a394935260e33cf8e32c2da5eaa0b2e12359da0e/crates/moonrun/src/main.rs#L320-L328

Rust binding is not provided for Instant.So even if he needed it, he couldn't use it. In init_env, instant now actually provides in __moonbit_time _unstable

lijunchen (approved):

lijunchen:
please clean up merge commit so we can merge it

PR #382, opened 2024-10-12
internal: enable --release flag for build | check | run | bundle

Young-Flash:
`build | check | run | bundle` is already in release mode by default (moonc run in release mode if no debug flag provided), this PR just remove the cli parsing error for --release flag for them

### DIFF
--- a/crates/moon/src/cli.rs
+++ b/crates/moon/src/cli.rs
@@ -139,6 +139,10 @@ pub struct BuildFlags {
     #[clap(long, short = 'g')]
     pub debug: bool,
 
+    /// Compile in release mode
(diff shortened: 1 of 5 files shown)


crates/moon/src/cli.rs
```diff
@@ -139,6 +139,10 @@ pub struct BuildFlags {
     #[clap(long, short = 'g')]
     pub debug: bool,
 
+    /// Compile in release mode
+    #[clap(long)]
+    pub release: bool,
```
lijunchen, line 144: set conflict

lijunchen (approved):

PR #384, opened 2024-10-12
internal: use cc -O2 for native --release

Young-Flash:
for `run | build | test` (which need link-core), use `cc -O2` if --release is provided for native backend

### DIFF
--- a/crates/moon/src/cli.rs
+++ b/crates/moon/src/cli.rs
@@ -64,7 +64,7 @@ use moonutil::{
     cli::UniversalFlags,
     common::{
         read_module_desc_file_in_dir, BuildPackageFlags, LinkCoreFlags, MooncOpt, OutputFormat,
-        SurfaceTarget, TargetBackend, MOONBITLANG_CORE, MOON_MOD_JSON,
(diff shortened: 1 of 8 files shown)


crates/moon/src/cli.rs
```diff
@@ -188,14 +188,43 @@ impl BuildFlags {
...
     let extra_build_opt = moon_mod.compile_flags.unwrap_or_default();
-    let extra_link_opt = moon_mod.link_flags.unwrap_or_default();
+    let mut extra_link_opt = moon_mod.link_flags.unwrap_or_default();
+
+    #[cfg(unix)]
+    match run_mode {
+        // need link-core for build, test and run
+        RunMode::Build | RunMode::Test | RunMode::Run => {
+            if build_flags.release && build_flags.target_backend == Some(TargetBackend::Native) {
+                // check if cc exists in PATH
```
lijunchen, line 209: FYI
```rust
#[test]
fn test_which_cc() {
    if let Ok(cc) = which::which("cc") {
        println!("{}", cc.display());
    }
}
```

lijunchen (approved):

PR #416, opened 2024-10-23
feat: moon check single package(and it's deps)

Young-Flash:


crates/moon/src/cli/check.rs
```diff
@@ -56,6 +56,17 @@ pub struct CheckSubcommand {
     /// Monitor the file system and automatically check files
     #[clap(long, short)]
     pub watch: bool,
+
+    /// The package(and it's deps) to check
+    pub package_path: Option<PathBuf>,
```
lijunchen, line 61: Why not the same as `moon test --package`?

lijunchen: Does this support fuzzy matching?

crates/moonbuild/src/gen/gen_check.rs
```diff
@@ -257,24 +265,63 @@ pub fn gen_check(
...
+
         if !pkg.wbtest_files.is_empty() {
-            let item = pkg_with_wbtest_to_check_item(
+            let mut item = pkg_with_wbtest_to_check_item(
                 &pkg.root_path,
                 m.get_all_packages(),
                 pkg,
                 moonc_opt,
             )?;
+            item.patch_file = patch_path_for_cur_pkg.clone();
```
lijunchen: Can we move this into `pkg_with_wbtest_to_check_item`?

crates/moonbuild/src/gen/gen_check.rs
```diff
@@ -257,24 +265,63 @@ pub fn gen_check(
     let _ = moonc_opt;
     let _ = moonbuild_opt;
     let mut dep_items = vec![];
-    for (_, pkg) in m.get_all_packages().iter() {
-        let item = pkg_to_check_item(&pkg.root_path, m.get_all_packages(), pkg, moonc_opt)?;
+
+    // --patch and --no-mi only apply over the package which specified by path
```
lijunchen: `--patch-file`?

crates/moonbuild/src/gen/gen_check.rs
```diff
@@ -257,24 +265,63 @@ pub fn gen_check(
...
+    let mut spec_pkg_to_check = None;
+    let mut patch_path = None;
+    let mut no_mi = false;
+
+    // if pkg is specified, check that pkg and it's deps; if no pkg specified, check all pkgs
+    let pkgs_to_check = match moonbuild_opt.check_opt.as_ref() {
```
lijunchen: The filter doesn't seem as clean as its counterpart in `gen_runtest`. The logic for checking should be simpler than the logic for testing.

crates/moon/src/cli/check.rs
Young-Flash: `moon test --package`  requires pkg name(username/hello/lib), while `package_path` for `check` requires the path of a pkg, which is more convient for ide extentsion 

> Does this support fuzzy matching?

no, this feat will not be used by user in general, it's used by IDE extentsion(AI)

crates/moonbuild/src/gen/gen_check.rs
Young-Flash: yes, but need to add extra args to `pkg_with_wbtest_to_check_item `

crates/moon/src/cli/check.rs
lijunchen: If so, I suggest adding a fancy flag for `package_path`

crates/moonbuild/src/gen/gen_check.rs
lijunchen: Or try to move some dirty work into `run_check_internal` to keep `gen_check` simple and clean.

PR #429, opened 2024-10-29
fix lock in pre-build

lijunchen:


crates/moonbuild/src/entry.rs
```diff
@@ -293,6 +293,12 @@ pub fn run_moon_pre_build(
     moonbuild_opt: &MoonbuildOpt,
     module: &ModuleDB,
 ) -> anyhow::Result<MoonPreBuildState> {
+    let common = moonbuild_opt.source_dir.join("target").join("common");
```
lijunchen: please help to polish this @Young-Flash

crates/moonbuild/src/pre_build.rs
```diff
@@ -118,13 +120,12 @@ pub fn load_moon_pre_build(
     }
 
     let mut hashed = n2graph::Hashes::default();
-    let common = moonbuild_opt.raw_target_dir.join("common");
-    if !common.exists() {
-        std::fs::create_dir_all(&common)?;
-    }
-    let _lock = FileLock::lock(&common)?;
+    let common = moonbuild_opt.source_dir.join("target").join("common");
```
lijunchen: and this

PR #444, opened 2024-11-04
feat: support patch for moon test

Young-Flash:
usage: moon test -p lib --patch /path/to/your/patch(_(wb)test).mbt


Young-Flash:
closed by https://github.com/moonbitlang/moon/pull/455

PR #490, opened 2024-11-25
feat: support moon install --bin

Young-Flash:
usage: 

- moon add `Young-Flash/test_publish`, it will add `Young-Flash/test_publish` into `./.mooncakes/`
- moon install --bin `Young-Flash/test_publish`, it will install pkg's link-core artifact(.wasm | .js | .exe)  if `is-main = true`(in moon.pkg.json) into `./.mooncakes/__moonbin__/`
    - if you want to compile to js backend, use moon install --bin `Young-Flash/test_publish` -- --target js
   - if you want to give a name to the bin which install into `./.mooncakes/__moonbin__/`, setting "bin-name" in moon.pkg.json
- in "pre-build" command, use `node $mooncake_bin/main.js`

cc @hackwaly


Young-Flash:
for lib author, a binary lib's moon.pkg.json:
```
{
  "is-main": true,
  "bin-target": "js", // optional, default to 'wasm-gc'
  "bin-name": "moonyacc" // optional, binary artifact name, default to package name
}
```

for user, `moon add Young-Flash/test_publish --bin`, moon.mod.json:
```
"bin-deps": {
  // simple foramt, install all main package
  "Young-Flash/test_publish": "0.2.0",
  // detailed foramt, local deps? specificy which main pkg to install? specify alias for main pkg?
  "username/flash": {
    "path": "/Users/flash/projects/my-project",
    "bin-pkg": [      // package name to install
      "main-js",
      {
        "name": "main-wasm",
        "alias": "m-wasm"
      }
    ]
  }
},
```

hackwaly:
works as expected

crates/moonutil/src/common.rs
```diff
@@ -58,6 +58,12 @@ pub const BLACKBOX_TEST_PATCH: &str = "_test.json";
 
 pub const MOON_DOC_TEST_POSTFIX: &str = "__moonbit_internal_doc_test";
 
+pub const MOON_BIN_DIR: &str = "__moonbin__";
+
+pub const MOONCAKE_BIN: &str = "$mooncake_bin";
+pub const MOD_DIR: &str = "$mod_dir";
+pub const PKG_DIR: &str = "$pkg_dir";
```
Young-Flash, line 65: cc @hackwaly

PR #491, opened 2024-11-25
internal: tweak passing cli args

Young-Flash:


crates/moonrun/src/sys_api.rs
```diff
@@ -47,8 +47,14 @@ fn construct_args_list<'s>(
...
+        // path of moonrun and the path of the wasm file
+        .take(2)
+        .chain(args.iter().cloned())
+        .collect();
+
     let arr = v8::Array::new(scope, args.len() as i32);
-    for (i, arg) in args.iter().enumerate() {
+    for (i, arg) in cli_args.iter().enumerate() {
         let arg = v8::String::new(scope, arg).unwrap();
         arr.set_index(scope, i as u32, arg.into());
```
lynzrand, line 59: I was debugging a commandline args passing issue to see why args passing in WASM backend contains the caller functions, and I found this (and f6eaefeca14b273b504d45f9be5978e758ab936e, too).

I remember this piece of code I originally wrote -- I even _collected_ the correct list of args that's not part of the options passed to `moonrun`. It's the part after the `--`. Why did you change this to the list of raw, unfiltered args that's not parsed and does not have a guaranteed sturcture? 

I can't think of any sane reason on making this change. If you need additional args to pass to the callee, you should've cange the args list to pass them directly, instead of just taking the whole freakin' list of raw args and stripping the first two (we don't even guarantee the user would pass the filename in the 2nd positon).

dude. just. why?

lynzrand: OK my miss. It's taking the first two args and passing the rest, but it's the same issue anyway. If you really need to pass `moonrun` path and the program being executed to your test runner, pass it _explicitly_, _in addition_ to the current args, instead of using the global `argv` here and breaking everything else.

lynzrand: Just taking a note here -- the correct way to make it behaving like it's wrapping something else does already exist in this codebase:

https://github.com/moonbitlang/moon/blob/28e91ab4770e80bf1ec189e1df79217b45d1fee4/crates/moon/src/cli/coverage.rs#L29-L43

PR #495, opened 2024-11-27
feat: allow config warn list by passing cli args

Young-Flash:


crates/moonrun/src/sys_api.rs
```diff
@@ -44,17 +44,10 @@ const INIT_SYS_API: &str = r#"
...
 ) -> v8::Local<'s, v8::Array> {
-    // align with node's behavior
-    // use this concat instead of std::env::args(), to ignore the possible --
-    let cli_args: Vec<String> = std::env::args()
-        // path of moonrun and the path of the wasm file
-        .take(2)
-        .chain(args.iter().cloned())
-        .collect();
-
+    let cli_args: Vec<String> = std::env::args().collect();
```
lynzrand, line 50: -- Continuing from https://github.com/moonbitlang/moon/pull/491/files#r2055327603, this diff is straight out code crime. _Nobody, ever,_ expected the raw list of args to **`moonrun`** being passed to user programs here.

PR #504, opened 2024-12-02
internal: tweak install bin

Young-Flash:
- remove alias in bin-deps[bin_pkg]
- change bin install location to bin_mod_path(use "node .mooncakes/hackwaly/moonyacc/moonyacc.js" in pre-build command)

cc @hackwaly


hackwaly:
it works.

PR #578, opened 2025-01-24
ci: use `ubuntu-latest` for coverage checks

rami3l:
Similar to https://github.com/moonbitlang/core/pull/1549, this should help mitigate our CI bottleneck caused by the shortage of macOS runners.


rami3l:
@lynzrand Can you reproduce this locally on Linux x64?

lynzrand:
> @lynzrand Can you reproduce this locally on Linux x64?

Can confirm the 8% coverage regress locally on x64-linux. Rerunning with LLVM...

rami3l:
It seems that `cargo-tarpaulin` will exit with an OOM error on our free CI runners. This is a known issue: https://github.com/xd009642/tarpaulin/issues/1639.

Issue #611, opened 2025-02-18
moon check && moon check should preserve warnings


bobzhang:
# Feature Request

When we run `moon check` for the first time, and `moon check` again the warning is lost

## Summary

For package a , and package b where a depends on b and a has warnings, we would prefer `moon check` will always report a's warnings while
b won't get rebuilt

## Checklist

- [x] I have added all necessary details to make it easy for the maintainers to understand my request.

PR #625, opened 2025-02-24
update n2 rev

lijunchen:


lijunchen:
Close https://github.com/moonbitlang/moon/issues/611

PR #648, opened 2025-03-06
feat: add llvm backend support

Young-Flash:


lynzrand (approved):
lgtm

crates/moon/src/cli/bundle.rs
```diff
@@ -71,7 +71,8 @@ pub fn run_bundle(cli: UniversalFlags, cmd: BundleSubcommand) -> anyhow::Result<
     // should move to `lower_surface_targets` when native backend being stable
     if cmd.all || cmd.build_flags.target == Some(vec![SurfaceTarget::All]) {
         targets.push(TargetBackend::Native);
-        if !cfg!(target_os = "windows") {
```
lynzrand: i think we can skip this and just let moonc crash if LLVM is not supported

lynzrand:
Problem is that, you don't know whether the `moonc` is compiled _with_ or _without_ LLVM support, just based on the target backend. It's totally valid to ship `moonc` without LLVM support on a platform that supports it. Thus, skipping _some_ platforms in `moon` don't make sense at all.

PR #740, opened 2025-04-10
feat: Report ambiguous imports

lynzrand:
Add feature to report ambiguous package name within project. Current logic:

- Errors out when any two package have the same full name.

The new code does/did not:

- Check whether the packages are actually used in an ambiguous place.
- Modify the import resolving logic. If one module's name is a path prefix of another, the import resolving might _still_ inaccurately report an unknown package, _same as the existing logic_.


lynzrand:
Went the other way to aim for a minimal change set to implement the feature. Currently reporting all ambiguous packages imported, regardless if they're acutally used ambiguously.

lijunchen (approved):

PR #769, opened 2025-04-23
fix: Obey env var `MOON_HOME` when looking for dirs

lynzrand:
Previously, `MOON_DIRS` does not obey the `MOON_HOME` env var set, but looks for `moonc` in PATH. 

This behavior is problematic if `moonc` is set as a shim or symbolic link, using the already-existing `home()` is a more robust option. For development cases where you need to find the files beside `moonc`, it might be better to just manually link them into `MOON_HOME`.

### DIFF
--- a/crates/moonutil/src/moon_dir.rs
+++ b/crates/moonutil/src/moon_dir.rs
@@ -31,10 +31,7 @@ pub struct MoonDirs {
 }
 
 pub static MOON_DIRS: std::sync::LazyLock<MoonDirs> = std::sync::LazyLock::new(|| {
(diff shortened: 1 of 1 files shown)


waterlens:
Will `moonc` be installed to a path like `/usr/local/bin/` or `/opt/bin/`? If so, it would be better to detect its presence in `home()` using `which` after finding that `MOON_HOME` is not set.

lynzrand:
I would say unlikely. And if in an unlikely event someone does we should just let them change the source code -- only supporting bundled version should be ok.
________________________________

Young-Flash (approved):

Issue #730, opened 2025-04-07
Unexpected File Extension in ELF Executable


ftcai:
Description:
When executing the command:
```console
moon run --target native linux.mbt
```
The generated artifact linux.exe in the target directory is identified as an ELF executable (confirmed via file command):
```console
linux.exe: ELF 64-bit LSB pie executable
```
- OS: WSL2 Debian Environment
- Moon Version: 0.1.20250401

Expected Behavior:
A standard ELF binary named 'linux' (without the .exe extension) to adhere to Linux naming conventions for executables.

hustcer:
I also get this suffix when building on macOS. It's quite strange—honestly, this is the first time I've seen it. I prefer a binary file without any extension.

lynzrand:
Thanks for the report! This is currently an expected behavior, following what the OCaml community usually do -- using `.exe` extension on all platforms, so no platform-specific behavior is needed. However, we acknowledge that such behavior is unexpected in some cases, and we will chat with our colleagues about what to do with it.

ftcai:
都191

ftcai:
- 都191

ftcai:
Sorry.My account issue.Typo fixed.

lynzrand:
No plan for changing the `.exe` extension in any foreseeable future. Closing this issue, feel free to reopen if it bothers you.

PR #775, opened 2025-04-26
fix(moonutil) *nix targets output executable files with .exe extension.

hanbings:
## Related Issues

- [x] Related issues: #730

## Type of Pull Request

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

  - When building on the Windows operating system, the output will be an executable file with a .exe extension, whereas on Linux, the output will be an executable file with a .out extension.
- [ ] No

- [x] No

## Checklist:


hanbings:
Just to add a small point: changing *.exe to *.out is mainly based on the convention that gcc and clang default to producing files with the .out suffix. 

When I searched online, most of the discussions I found were about compatibility. What I would like to ask is: since moonbit is a new language, is it necessary to carry over such **historical conventions**?
Perhaps it could follow an approach similar to Rust (apologies for bringing up another language here, but I thought it would be an appropriate example), and by default, output an ELF file without any suffix on *nix systems?

cc @ftcai

waterlens:
Use `.exe` extensions makes it possible to have a consistent binary name among all OSs we supported. It's not something just about compatibility.
Is there a specific case where users need a file name without the extension, otherwise something won't work?

By the way, removing on Unix `.exe` can lead to many duplicated test cases, which is not ideal for maintainability:
`https://github.com/moonbitlang/moon/blob/df8b155ac39349359c71624633eb62581d5edac7/crates/moon/tests/test_cases/mod.rs`

hanbings:
> Is there a specific case where users need a file name without the extension, otherwise something won't work?

I don't think there are any programs that actually rely on the file extension to work; it's just that its behavior on *nix systems is a bit unusual. :)

If this is not a bug but an intended feature, then I think we can go ahead and close this PR.

Thanks.

Issue #801, opened 2025-05-09
Tracking issue for pre-build configuration scripts


lynzrand:

illusory0x0:
- #427

lynzrand:
> * [support `post-build` #427](https://github.com/moonbitlang/moon/issues/427)

We currently have 2 prebuilds, might need an overhaul on that before moving to post-build and stuff. Maybe we need an overhaul on the build system as a whole (I'm not yet sure about how to do that yet). Anyway, comments are welcomed!

lynzrand:
Still need overhaul.

PR #803, opened 2025-05-09
Pre-build config script MVP

lynzrand:
## Related Issues

- [x] Related issues: #801 

## Type of Pull Request

- [ ] Bug fix
- [x] New feature
    - [x] I have already discussed this feature with the maintainers.
- [ ] Refactor
- [ ] Documentation

  - A new flag `--moonbit-unstable-prebuild` is added to `moon.mod.json`, which must point to a JS file if set. If set, the script will be evaluated (currently unconditionally) using `node` before all builds start. It will generate configurations that changes build behavior.
  - In link flags in `moon.pkg.json`, `${build.<var>}` will be replaced by the corresponding variables in build script.
  - Build script will be able to emit link configs, which will then be applied to all dependents of the specified package.
  - See `crates/moonutil/src/build_script.rs` for the input and output data sturcture design
- [ ] No

- [x] No

## Checklist:

- [ ] Compatible with Windows/Linux/macOS (the build scripts themselves should run in a target dependant way)
- [x] Add a `build.py` support


lynzrand:
@Young-Flash Is this an invariant not holding? 

Single file builds doesn't seem to be populating the graph.

https://github.com/moonbitlang/moon/actions/runs/14964995416/job/42033245096?pr=803#step:15:263

```
[crates/moonutil/src/module.rs:374:9] &self.graph = Graph {
    Ty: "Directed",
    node_count: 0,
    edge_count: 0,
    node weights: {},
    edge weights: {},
}

thread 'main' panicked at /home/rynco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/fixedbitset-0.4.2/src/lib.rs:225:9:
put at index 0 exceeds fixbitset size 0
stack backtrace:
   0: rust_begin_unwind
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_fmt
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/panicking.rs:72:14
   2: fixedbitset::FixedBitSet::put
             at /home/rynco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/fixedbitset-0.4.2/src/lib.rs:225:9
   3: <fixedbitset::FixedBitSet as petgraph::visit::VisitMap<Ix>>::visit
             at /home/rynco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/petgraph-0.6.5/src/visit/mod.rs:408:10
   4: petgraph::visit::traversal::Dfs<N,VM>::next
             at /home/rynco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/petgraph-0.6.5/src/visit/traversal.rs:110:16
   5: <petgraph::visit::traversal::Dfs<<G as petgraph::visit::GraphBase>::NodeId,<G as petgraph::visit::Visitable>::Map> as petgraph::visit::traversal::Walker<G>>::walk_next
             at /home/rynco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/petgraph-0.6.5/src/visit/traversal.rs:504:9
   6: <petgraph::visit::traversal::WalkerIter<W,C> as core::iter::traits::iterator::Iterator>::next
             at /home/rynco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/petgraph-0.6.5/src/visit/traversal.rs:484:9
   7: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/iter/adapters/map.rs:108:9
   8: moonbuild::gen::util::calc_link_args
             at /home/rynco/Projects/moonbuild/crates/moonbuild/src/gen/util.rs:134:16
   9: moonbuild::gen::gen_runtest::gen_link_blackbox_test
             at /home/rynco/Projects/moonbuild/crates/moonbuild/src/gen/gen_runtest.rs:809:20
  10: moonbuild::gen::gen_runtest::gen_runtest
             at /home/rynco/Projects/moonbuild/crates/moonbuild/src/gen/gen_runtest.rs:957:37
  11: moonbuild::runtest::load_moon_proj
             at /home/rynco/Projects/moonbuild/crates/moonbuild/src/runtest.rs:43:20
```

Young-Flash:
yeah, single file build(moon run single .mbt file which no belong to any package) hasn't involve n2, it's using subprocess to 'build-package', 'lnik-core', and run executable

lynzrand:
That run is acutally a single-file test, which seems to be using n2 and ModuleDB. However, the DB is not populated correctly to hold its usual invariants. Gonna fix that here orelse our new impl won't work.

lynzrand:
@Young-Flash The new behavior only links in the target backend if the corresponding key in `link` exists, like `"link": { "wasm-gc": {} }` will link on wasm-gc while `"link": {}` won't. 

Is this still expected, or should we only apply this behavior to the native backend (i.e. native links iff `"link": true` or `"link": {"native": { something that's not stub linking } }`, but other backends links only checking whether `link` is not `null`)?

Young-Flash:
how about extract `native` filed out of `link`? my thought: for wasm wasm-gc js backend, `moonc link-core` emit the final artifact which could be executed, and `wasm`, `wasm-gc`, `js` config in `link` are passing to `moonc link-core`; while for `native` backend, `moonc link-core` will not emit the final exe now. in this way, we keep the `link` logic for other backend apart from `native`

lynzrand:
> in this way, we keep the `link` logic for other backend apart from `native`

I don't think that's consistent. If you do care about `link`'s purity, maybe we should split the params for C stubs out instead.

lynzrand:
After discussion: We will move stub-related config out into another config key, presumably named `link-stub` or similar. Then, `link` will contain purely link-related things.

Current config style will still be supported, i.e. the package will be linked into an executable/library if the `link` key is present. But if the user is using one of the native backends, the criteria will be `link.native` contains a key that is _not_ related to C-stubs.

We should warn the user on the presence of stub-related keys in `link.native` and recommend them to migrate, while doing the migration when building. Once the new key is widely adopted, we drop support for stubs in `link.native`.

The migration will be done in another PR. This PR will retain the current config.

crates/moonbuild/src/build_script.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+        std::borrow::Cow::Owned(new_s) => {
+            *s = new_s;
+        }
+    }
+    Ok(())
+}
+
+static NODE_EXECUTABLE: LazyLock<Option<&str>> = LazyLock::new(|| {
+    if which::which("node").is_ok() {
```
Young-Flash: for our own ide in windows, we need `node.cmd`

cc @bzy-debug

https://github.com/moonbitlang/moon/blob/b27a09861027b74d64453eb359cc2bd916ce7f4f/crates/moonbuild/src/build.rs#L46-L50

crates/moonutil/src/module.rs
```diff
@@ -366,6 +367,19 @@ impl ModuleDB {
             }
         })
     }
+
+    /// Get an iterator returning all dependencies of a certain package. The
+    /// order is unspecified.
+    pub fn all_deps_of_pkg(&self, pkg: &str) -> Option<impl Iterator<Item = &Package>> {
```
Young-Flash: we have `get_filtered_packages_and_its_deps_by_pkgname`, can we use it here?

Young-Flash (approved):
mostly LGTM

crates/moon/tests/test_cases/prebuild_config_script/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+    let found_c_flags_replacement = OnceCell::<()>::new();
+    let found_link_flags = OnceCell::<()>::new();
+    for line in lines {
+        if line.contains("stub.c") {
+            found_c_flags_replacement
+                .set(())
+                .expect("c stub compilation found twice");
+            assert!(line.contains("-D HELLO=------this-is-added-by-config-script------"));
+        }
+        if line.contains("cc -o ./target/native/release/build/main/main") {
```
lynzrand: TODO: Needs to supply args CL expects

crates/moonutil/src/module.rs
lynzrand: yeah might be. i wasn't sure about the semantics of that function so i didn't use it here, but it should work

lynzrand:
Will need an autosquash rebase before we merge; this is just to see how well the bugs are fixed

Young-Flash (approved):
good job!

PR #839, opened 2025-05-30
Fix prebuild config not having effect in tests

lynzrand:
### Changes

- Add link flags for native compilation for test files.

### TODO

- [x] ~~`tcc` or `tcc -run` don't work with the given args on my machine, while system `cc` works~~ Seems like a problem on my machine instead of tcc itself
- [x] Does not work on macOS

### DIFF
--- a/crates/moonbuild/src/gen/gen_build.rs
+++ b/crates/moonbuild/src/gen/gen_build.rs
@@ -1135,7 +1135,7 @@ pub fn gen_link_stub_to_dynamic_lib_command(
     // TODO: There's too many kinds of flags, need to document what each one do
     let cc_flags = native_stub_cc_link_flags
         .into_iter()
-        .chain(native_cc_link_flags.into_iter())
+        .chain(native_cc_link_flags)
         .collect::<Vec<_>>();
 
     let shared_runtime_dir = Some(runtime_path.parent().unwrap());


Young-Flash (approved):

lynzrand:
@waterlens is this behavior still right? 

https://github.com/moonbitlang/moon/actions/runs/15409898059/job/43359617907?pr=839#step:16:291

crates/moonbuild/src/gen/gen_build.rs
```diff
@@ -1121,16 +1121,31 @@ pub fn gen_link_stub_to_dynamic_lib_command(
...
+        .into_iter()
+        .chain(native_cc_flags.into_iter())
+        .chain(native_cc_link_flags.into_iter())
+        .collect::<Vec<_>>();
+
     let shared_runtime_dir = Some(runtime_path.parent().unwrap());
     let cc_cmd = make_linker_command::<_, &Path>(
         CC::default(),
-        native_stub_cc.map(|cc| {
+        native_cc.map(|cc| {
```
waterlens: Why change from `native_stub_cc` to `native_cc`?

waterlens: If we have compiled `some_stub.c` into `some_stub.o` with `stub_cc`, it seems more reasonable to use the same compiler(driver) to link it into `libsome_stub.so`.

lynzrand: IIRC this links every library we have built until now, along with the runtime and many things. We are even feeding the CC link flags into here, so maybe it makes sense to use the `native_cc`?

waterlens: I think `gen_link_stub_to_dynamic_lib_command` will only be called when using `tcc -run`?https://github.com/moonbitlang/moon/blob/db27b817d866182ededfa6505a4cdf0aea92850e/crates/moonbuild/src/gen/gen_build.rs#L1536-L1541
In this scenario, `cc` will be `tcc` and the `stub-cc` will be one of the normal C compilers. `tcc` cannot link object files produced by the other `stub-cc`.

lynzrand: Yes, it will only be used in `tcc -run`. I am proposing this because with external libraries `tcc` can't read many C compiler flags, and this compilation to a dylib will complain about unknown symbols, and thus the C linking flags need to be passed here.

So maybe yes, we should use `stub_cc` here.

waterlens: We can keep the changes related to `cc_flags` but still use `stub_cc`. Anyway, currently we (and MoonBit users) mixed the use of `cc-link-flags`, `cc-flags`, `stub-cc-flags`, `stub-cc-link-flags`, which is bad :(

lynzrand: legit. Will change.

lynzrand: Just reverted back to stub cc. Is the flag usages correct, or should we remove the `native_cc_flags` here? I don't think I know the process good enough now :(

waterlens: The code LGTM. Yeah, I shall document them if I have time as these options are complicated. I think their names have suggested their purposes well, but with one **exception**: `stub-cc-link-flags` won't be used unless we are at `tcc -run` mode. This exception occurs because, in normal modes, we do not have a separate stage for linking stubs. Instead, stubs and object files compiled from `.mbt` files are linked simultaneously, where the `cc-link-flags` apply.

lynzrand: Will you approve this change though? The docs can be added when we're refactoring moonbuild and documenting existing behavior.

crates/moonbuild/src/gen/gen_build.rs
```diff
@@ -1127,6 +1127,22 @@ pub fn gen_link_stub_to_dynamic_lib_command(
...
+        .unwrap_or_default();
+    let native_cc_link_flags = item
+        .native_cc_link_flags(moonc_opt.link_opt.target_backend)
+        .map(|it| it.split(" ").collect::<Vec<_>>())
+        .unwrap_or_default();
+
+    // TODO: There's too many kinds of flags, need to document what each one do
+    let cc_flags = native_stub_cc_link_flags
+        .into_iter()
+        .chain(native_cc_flags.into_iter())
```
waterlens: I am not sure if it would be appropriate to append `cc_flags` here. For me it only mitigates the following problem:
Users always misuse C compiler related flags, for example:
```
-I/some/dir: only compiler flags
-lsomelib, -Wl,some_option: only linker flags
-fsanitize=address, -static: both compiler flags and linker flags
```
They always write the options in only one place, when they should appear in both `cc-flags` and `cc-link-flags`. Then they complain `moon` doesn't pass the arguments to the compiler :(

lynzrand: Then I think we can remove this? cc flags should be only for the stubs?

Actually I question the idea of making C first class citizen in the build system but anyway... I will remove this and see.

lynzrand:
Tested on Linux & Mac with python.mbt, should be good to go.

PR #850, opened 2025-06-04
fix: moon run single mbt for llvm backend

Young-Flash:


lynzrand (commented):
reviewed offline. works.

Issue #847, opened 2025-06-04
LLVM backend: build failed for single file HelloWorld


EricTianC:
# Bug Report

### OS

Operating System: Windows

### MoonBit CLI Tools Version

```
moon 0.1.20250529 (daaecb6 2025-05-29) ~\.moon\bin\moon.exe
moonc v0.1.20250529+8a98c8e02 ~\.moon\bin\moonc.exe
moonrun 0.1.20250529 (daaecb6 2025-05-29) ~\.moon\bin\moonrun.exe
```

## Steps to Reproduce

write the helloworld code into `hello_moonbit.mbt` file

```moonbit
fn main {
  println("Bye llvm")
}
```

execute `moon run .\hello_moonbit.mbt --target=llvm` in the same directory

## Expected Behavior

output `Bye llvm`

## Actual Behavior

```
moon\run\single\hello_moonbit.mbt:2:3-2:10 [E4021] The value identifier println is unbound.
error: failed to run: moonc build-package D:\Archives\moonbit_projects\casuals\hello_moonbit.mbt -o D:\Archives\moonbit_projects\casuals\target\hello_moonbit.core -std-path C:\Users\EricT\.moon\lib\core\target\llvm\release\bundle -is-main -pkg moon/run/single -g -O0 -source-map -target llvm 
```

## Checklist
- [x] (Optional) My case is minimal enough to be reproducible.

lynzrand:
Does the same issue exist in Native (C) backend, aka `--target native`?

lynzrand:
My guess is that core library bundling has not been done for the LLVM backend. This is a known issue of the installation script and is being fixed.

In the meantime, you can run `moon -C ~/.moon/lib/core bundle --target llvm` and retry your run.

If you still encounter a bug saying something like "unknown symbol `__main`", that is a separate issue and is also getting fixed actively.

PR #851, opened 2025-06-04
fix: moon run single mbt should set -llvm-target for msvc

Young-Flash:
close https://github.com/moonbitlang/moon/issues/847


lynzrand (approved):
lgtm

PR #855, opened 2025-06-05
[Rupes Recta] Document about the behavior of the compile

lynzrand:
- **docs: Add RFC for project Rupes Recta**
- **docs: Add documentation for MoonBit package compilation**
- **docs: Add documentation of how a package is built**
- **docs: Add documentation about virtual packages**


lynzrand:
Maybe we should merge this (although with WIP docs) first? This thing is documentation-only, so it will not interfere with existing code.

Young-Flash:
ok

Issue #802, opened 2025-05-09
Refactor and cleanup of `moon` codebase to reduce clutter and code bloat


lynzrand:
We have been weighting deliver speed over codebase clarity, but we will need to do major refactors more often to keep the codebase clean. The process will probably begin with a major refactor of the `moonbuild` crate, but details are still TBD.

# Plans

TBD

lynzrand:
We now have the first feature parity checkbox set at `moon fmt`!

lynzrand:
The second one: `moon doc`

lynzrand:
TBD: should we unify the behavior of stripping and debuginfo?

lynzrand:
After a lengthy rewrite of 6 whole months, including about half of the time in a long tail making sure the new backend behaves exactly like the old one, I now declare that the rewrite has reached feature parity.'

The rewrite consists of over 300 PRs, over 1000 commits, and over 50000 lines of changes. Thanks everybody for the heavy work and cooperation!

20 tests are marked as _won't fix_ and will be promoted once the switch is flipped. They are being logged in `xtask/rr_wont_fix.txt`, and the following is a copy of it:

```
# RR tests that won't be fixed (known differences)
# One fully-qualified test name per line.
# Lines starting with # and blank lines are ignored.

# error message differs
test_cases::warns::test_warn_list_dry_run
test_cases::test_nonexistent_package
test_cases::test_validate_import
test_cases::fuzzy_matching::test_fuzzy_matching
test_cases::test_specify_source_dir_with_deps_002

# deprecated --build-graph option
test_cases::moon_build_package::test_moon_build_package

# legacy implementation is buggy
test_cases::cond_comp::test_cond_comp

# rr improves the behavio: when `moon test main` rr will not check dependencies' white/blackbox test
test_cases::filter_by_path::test_moon_check_filter_by_path_success

# Legacy includes irrelevant tasks while RR doesn't
test_cases::test_moon_check_filter_package
test_cases::test_moon_check_package_with_patch
test_cases::test_moon_doc_dry_run
test_cases::native_backend::cc_flags::test_native_backend_cc_flags_with_env_override
test_cases::filter_by_path::test_moon_check_filter_by_path_failure
test_cases::filter_by_path::test_moon_check_filter_by_path_success

# tcc -run execution method difference
test_cases::native_backend::tcc_run::test_native_backend_tcc_run
test_cases::targets::test_many_targets

# test order different
test_cases::test_virtual_pkg

[...]

PR #864, opened 2025-06-09
Change `Rc` to `Arc` & use `SlotMap` in module resolver

lynzrand:
This PR replace the usage of `Rc` with `Arc` in mooncakes resolver, and uses `SlotMap` instead of a hand-crafted slotmap-like structure based on `IndexSet`. The target of this change is to aid multithreaded building later into the rewrite process. You may skip this PR for merging for now, if you want.

The dependency solver and registry part is still single-threaded, and we have no plan in the near future on making them multithreaded.

## Drawbacks

Yeah, we aren't even removing items from `SlotMap`s, so half of the type is mostly useless. However, the provided `SecondayMap` makes structures that use the sequential `ModuleId` as key far easier to create and use. I hope that justifies the change.

## Related Issues

- [x] Related issues: #802 rewrite of `moon`

## Type of Pull Request

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation

  - ____
- [x] No

- [ ] No
- [x] Yes: Added `slotmap`, which is very small (and even so when we aren't using its higher-level features)

## Checklist:

### DIFF
--- a/crates/moonbuild/src/build_script.rs
+++ b/crates/moonbuild/src/build_script.rs
@@ -22,7 +22,7 @@
 use std::{
     collections::HashMap,
     io::Write,
-    path::{Path, PathBuf},
+    path::Path,
     process::{Command, Stdio},
     str::FromStr,
 };


crates/mooncake/src/resolver/mvs.rs
```diff
@@ -476,7 +477,12 @@ mod test {
         let mut env = ResolverEnv::new(&registry);
         let result = resolver.resolve(&mut env, &roots).expect("Resolve failed");
 
-        let id = result.id_from_mod_name(&root_ms).unwrap();
+        let id = result
+            .all_modules_and_id()
+            .filter(|(_, ms)| ms.name == module_name && ms.version == version)
```
rami3l: Nit: `.find_map()`

crates/mooncake/src/resolver/mvs.rs
```diff
@@ -589,11 +595,20 @@ mod test {
         ));
     }
 
+    fn id_from_mod_name(result: &ResolvedEnv, mod_name: &ModuleName) -> Option<ModuleId> {
+        result
+            .all_modules_and_id()
+            .find(|(_, m)| m.name == *mod_name)
```
rami3l: Nit: `.find_map()`

rami3l (approved):

crates/mooncake/src/resolver/mvs.rs
lynzrand: I think .find(...).map(...) is clearer?

PR #867, opened 2025-06-10
Add preferred target configuration for modules

### LINKED ISSUE #865: Preferred target for modules
# Feature Request

## Summary

Add a field in `moon.mod.json` that tells the toolchain the preferred target backend of the given module. 

## Motivation

Some modules are made for a specific target backend, they should not use the default `wasm-gc`

## Detailed Description

Add `$.preferred-target` field in `moon.mod.json`, whose value is either one of the backends or `null`. When running `moon test`, `moon build` and other toolchain commands, if the value is set, we should use its value as the default backend to run the toolchain (as if `--target=...` was set). Otherwise, we use the default backend, which is currently `wasm-gc`.

## Checklist

- [ ] I have added all necessary details to make it easy for the maintainers to understand my request.

lynzrand:
## Related Issues

- [x] Related issues: #865 

closes #865 

## Type of Pull Request

- [ ] Bug fix
- [x] New feature
    - [x] I have already discussed this feature with the maintainers.
- [ ] Refactor
- [ ] Documentation

  - A new field `preferred-target` is added to `moon.mod.json`. Setting this field will cause all `moon ____` build commands to use this target as the default target to run, when no other overrides are provided.
- [ ] No

- [x] No

## Checklist:

### DIFF
--- a/crates/moon/src/cli.rs
+++ b/crates/moon/src/cli.rs
@@ -247,7 +247,10 @@ pub fn get_compiler_flags(src_dir: &Path, build_flags: &BuildFlags) -> anyhow::R
         OutputFormat::Wasm
     };
 
-    let target_backend = build_flags.target_backend.unwrap_or_default();
+    let target_backend = build_flags
+        .target_backend
+        .or(moon_mod.preferred_target) // if we have specified in module config
+        .unwrap_or_default();
 
     if target_backend == TargetBackend::Js && output_format == OutputFormat::Wat {
         bail!("--output-wat is not supported for --target js");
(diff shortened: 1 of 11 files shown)


Young-Flash:
should we remove `supported-targets` in `moon.pkg.json`?

lynzrand:
> should we remove `supported-targets` in `moon.pkg.json`?

AFAICT, not yet? We will need to assess the impact and do migrations and stuff...

But I do think that we need a discussion on which features are needed and which are not for MoonBuild, it feels like there are two ways to do everything in this system.

(BTW when are we updating the test file contents, they're failing)

Young-Flash:
ok

I will update the outdate syntax content in test file. https://github.com/moonbitlang/moon/pull/868

Young-Flash (approved):
LGTM

Issue #861, opened 2025-06-06
can not build  hello on window LINK : fatal error LNK1181: m.lib


liyunde:
# Bug Report

## Environment
MSVC\14.43.34808

clang -v
clang version 20.1.0
Target: x86_64-pc-windows-msvc
Thread model: posix

### OS

Operating System: __Windows11__

### MoonBit CLI Tools Version

moon 0.1.20250529 (daaecb6 2025-05-29) ~\.moon\bin\moon.exe
moonc v0.1.20250529+8a98c8e02 ~\.moon\bin\moonc.exe
moonrun 0.1.20250529 (daaecb6 2025-05-29) ~\.moon\bin\moonrun.exe
```

## Steps to Reproduce

moon new hello
other use default

moon run .\src\main\ is normal

moon build --release  --target native

## Expected Behavior

can gen main.exe

## Actual Behavior

LINK : fatal error LNK1181: 无法打开输入文件“m.lib”
clang: error: linker command failed with exit code 1181 (use -v to see invocation)
error: failed to run build for target Native

Caused by:
    failed when building

## Checklist
- [ ] (Optional) My case is minimal enough to be reproducible.

Young-Flash:
how about use `msvc`?

liyunde:
moon version --all
moon 0.1.20250618 (ecea0c0 2025-06-18) ~\.moon\bin\moon.exe
moonc v0.6.18+06c89576e ~\.moon\bin\moonc.exe
moonrun 0.1.20250618 (ecea0c0 2025-06-18) ~\.moon\bin\moonrun.exe

After updating to the latest version, the error still occurs consistently.

This error is reproducible in environments where both the Windows SDK and Visual Studio are present.

liyunde:
An unnecessary -lm flag was added to the build command. This flag is used to link the math library on Linux systems, but it is not required or supported on Windows.

Guest0x0:
Is `cl.exe` in your `$PATH`?

liyunde:
> Is `cl.exe` in your `$PATH`?
@Guest0x0 
PS D:\> cl
用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.44.35213 版
版权所有(C) Microsoft Corporation。保留所有权利。

The error occurs only in the PowerShell MSVC environment; it executes normally in CMD, and also works correctly in the GCC environment.

[...]

PR #944, opened 2025-07-24
fix: remove `-lm` flag for windows native target

CrazyboyQCD:
## Related Issues

- [x] Related issues: #861

## Type of Pull Request

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

  - Remove `-lm` flag for building `msvc` target
- [ ] No

- [x] No

## Checklist:

- [] Tests added/updated for bug fixes or new features
- [] Compatible with Windows/Linux/macOS

### DIFF
--- a/crates/moonutil/src/compiler_flags.rs
+++ b/crates/moonutil/src/compiler_flags.rs
@@ -461,7 +461,7 @@ where
 
     // Link against some common libraries
     if cc.is_gcc_like() {
-        if cc.is_full_featured_gcc_like() {
+        if !cfg!(target_os = "windows") && cc.is_full_featured_gcc_like() {
             buf.push("-lm".to_string());
         }
         if let Some(dyn_lib_path) = config.link_shared_runtime.as_ref() {
@@ -655,7 +655,10 @@ where
     buf.extend(src.iter().map(|s| s.as_ref().to_string()));
 
     // Link against some common libraries
-    if cc.is_full_featured_gcc_like() && config.output_ty != OutputType::Object {
+    if !cfg!(target_os = "windows")
+        && cc.is_full_featured_gcc_like()
+        && config.output_ty != OutputType::Object
+    {
         buf.push("-lm".to_string());
     }
 


lynzrand:
Thanks for your PR! 

We're sorry about the current condition of our codebase that made things unclear, but the implementation in this PR is a little off. The distinction of MSVC and GCC needs to be done at *runtime* instead of *compile-time*, so `cfg!` macros don't work here. You can check the following code for reference on how to detect MSVC while we're sorting things out. 

https://github.com/moonbitlang/moon/blob/b897b6d3d28f629b793f555c106c6f97f8748487/crates/moonbuild/src/gen/gen_build.rs#L743-L749

CrazyboyQCD:
> Thanks for your PR!
> 
> We're sorry about the current condition of our codebase that made things unclear, but the implementation in this PR is a little off. The distinction of MSVC and GCC needs to be done at _runtime_ instead of _compile-time_, so `cfg!` macros don't work here. You can check the following code for reference on how to detect MSVC while we're sorting things out.
> 
> https://github.com/moonbitlang/moon/blob/b897b6d3d28f629b793f555c106c6f97f8748487/crates/moonbuild/src/gen/gen_build.rs#L743-L749

I test that `clang` also fails to link `m.lib`, and as comment in https://github.com/moonbitlang/moon/issues/861#issuecomment-2987207774 said, `windows` doesn't need this flag not just for `msvc` target.

lynzrand:
Okay, I have just verified that both MSVC and GNU targets on Windows have math functions defined in MSVCRT. We will discuss this change internally before proceeding.

lynzrand:
At the meantime, you might want to sign the CLA to make your changes mergable :)

Check this CI for more details: https://github.com/moonbitlang/moon/actions/runs/16511697077/job/46694762934?pr=944

CrazyboyQCD:
> At the meantime, you might want to sign the CLA to make your changes mergable :)
> 
> Check this CI for more details: https://github.com/moonbitlang/moon/actions/runs/16511697077/job/46694762934?pr=944

Something wrong with the signing page https://github.com/moonbitlang/website/issues/2.

lynzrand (approved):

lynzrand:
Do note that if we're doing cross-compilation in the future, we can't use a compile-time `cfg!` as condition.

peter-jerry-ye:
Thanks again for the patch, and sorry for the mixed signals here.

A previous maintainer comment suggested this had effectively been addressed, but the PR itself was never actually closed. We’re cleaning that up now.

We’re closing this PR as superseded by #1586.

The bug we were able to confirm is the MSVC-family case, where passing `-lm` is incorrect and can fail at link time. The merged fix in #1586 handles that by deciding from the selected compiler target at runtime, which is a better fit for how Moon models C toolchains here.

This PR used a host-OS check (`cfg!(target_os = "windows")`), which is broader than the specific bug we could verify, so we’re treating #1586 as the final fix for this area.

Thanks again for the report and the patch.

PR #982, opened 2025-08-28
fix: Fix param order in render callback

lynzrand:


lynzrand:
This is an extremely easy bugfix for the issue at https://github.com/moonbitlang/core/actions/runs/17286608035/job/49064933134?pr=2607, so I bypassed the reviewing

PR #995, opened 2025-09-03
chore: Ignore generated files in test dir

lynzrand:
This PR is a spinoff of #802's main progress.

This PR ignores the files that might be created during `moon check`ing of the test case projects. The `moon check` might be triggered by AI coding tools, causing clutter in the repo that should not be comitted.


lynzrand:
Simple PR and passed AI review. Merged.

PR #1023, opened 2025-09-16
simplify test driver

Guest0x0:
This PR clean up the test driver mess, utilize the `#cfg` attribute recently introduced to avoid duplicated code in test drivers. The new test driver architecture is as follows:

- `common.mbt`: things shared by all driver variants, such as input parsing, entry point
- `bench_driver_template.mbt`: driver for `moon bench`
- `test_driver_template.mbt`: test driver for `moon test` of everything, including `moonbitlang/core/builtin`. When all tests have no argument, we don't want the driver to introduce unnecessary dependency on `moonbitlang/core/test`. To solve this problem, the type of test argument is turned into a type alias `Moonbit_Test_Driver_Internal_Test_Arg`, with a function `moonbit_test_driver_internal_new_test_arg` to create test arguments, and then:
    - when all tests has no argument, `no_args_template.mbt` will be added to the driver, which use a dummy type and dummy function for test argument, eliminating the dependency on `moonbitlang/core/test`
    - when test argument is used, `with_args_template.mbt` is used, which uses `@test.T` as test argument

In this new architecture, there is no duplicated code, making maintenance easier. This PR would also make the future introduction of `async test` easier. 

## Type of Pull Request

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Documentation

  - ____
- [X] No

- [x] No

## Checklist:

- [X] Tests added/updated for bug fixes or new features
- [X] Compatible with Windows/Linux/macOS


crates/moon/src/cli/generate_test_driver.rs
```diff
@@ -47,7 +47,7 @@ pub struct GenerateTestDriverSubcommand {
 
     /// The target backend for the generated test driver.
     #[clap(long = "target")]
-    target_backend: TargetBackend,
+    _target_backend: TargetBackend,
```
Guest0x0, line 50: I am not familiar with the main `moon` code base, so I leave the command line switch as-is here. Should be removed in the future

lynzrand: that's fine. maybe leave a comment and i will remove it later.

PR #1057, opened 2025-09-29
fix unused link path when compiling with non-tcc

Guest0x0:
Currently, native backend will unconditionally pass the flag `-L$MOON_HOME/lib` to `cc`. However, this link flag is for `libtcc1.a`, which is only needed by `tcc`. Indeed, `clang` will currently complain about unused link flag on basically every MoonBit project.

This PR fixes the warning by passing the `-L$MOON_HOME/lib` flag only when `cc` is `tcc`. In addition to tests in this repo, I locally tested the change on `moonbitlang/core` and `moonbitlang/async`, and everything works well.


Guest0x0:
update: there is one missing place with `-L$MOON_HOME/lib` when linking C stub

PR #1060, opened 2025-09-30
[Rupes Recta] Fix and stabilize build commands, enhance debugging

### LINKED ISSUE #1059: [RR] Virtual package addition broke build graph
Declaration of output file is wrong in `moon check`.

lynzrand:
This PR enhances the stability of the command arguments and ordering for Rupes Recta:

- The package order in `moonc link-core` sorted.
- The metadata output order is sorted.
- The input files list of each build command is sorted.

With these stabilizations, `n2` should no longer rebuild if e.g. internal hashmap iterator order changes and cause the aforementioned iteration order to change.

This PR fixes a bug that causes virtual packages to fail to build in `moon check`, due to the corresponding `moonc check` command for a virtual package with default implementation should not announce the `.mi` interface as its output (and as an input instead).

It also fixes when virtual packages are always rebuilt in `moon build`, due to a similar bug: the `-no-mi` argument is not emitted, so the output of that build overwrites the original input `.mi` file.

Fixes #1059 

## Related Issues

- [x] Related issues: #802, #1059 

## Type of Pull Request

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

- [ ] No

- [x] No

## Checklist:


lynzrand:
cc @bobzhang This PR should solve the issue observed before when building core.

lynzrand:
We will also need a way to properly test the new build system, or similar bugs will also surface in the future. AFAICT there are a lot of false failures in the current test suite that we can't tell unless we read each one carefully. The sheer number of failed tests (~95/~290, about 1/3) makes it unfeasible to fix in batches.

lynzrand:
Self-reviewed and fixed a bunch of other things.

crates/moonbuild-rupes-recta/src/build_plan/builders.rs
```diff
@@ -626,12 +644,20 @@ impl<'a> BuildPlanConstructor<'a> {
...
                     if !seen.contains(&child) {
                         seen.insert(child);
                         stack.push((child, false));
                     }
                 }
+                // Stable sort the newly added children
+                stack[stack_before_children..].sort_by_key(|(t, _)| {
+                    let pkg_name = &self.input.pkg_dirs.get_package(t.package).fqn;
+                    (pkg_name, t.kind)
+                });
```
lynzrand, line 660: I'm unsure if this might hurt perf, but it's needed for a stable topo order.

PR #1076, opened 2025-10-11
[Rupes Recta] Implement watch mode

lynzrand:
Closes #1064 

This PR implements watch mode for the new `moon` backend. It also refactors the original watch mode code to decouple it from the legacy code sturctures.

# Related Issues

- [x] Related issues: #1064,#802

## Type of Pull Request

- [ ] Bug fix
- [x] New feature
    - [ ] I have already discussed this feature with the maintainers.
- [ ] Refactor
- [ ] Performance optimization
- [ ] Documentation
- [ ] Other (please describe):

## Does this PR change existing behavior?

- [x] Yes (please describe the changes below)
  - ____
- [ ] No

## Does this PR introduce new dependencies?

- [x] No
- [ ] Yes (please check binary size changes)

## Checklist:

- [ ] Tests added/updated for bug fixes or new features
- [ ] Compatible with Windows/Linux/macOS

PR #1110, opened 2025-10-21
Use naive watch mode for legacy code path

### LINKED ISSUE #1106: current `moon check -w` is fragile
# Bug Report
when I remove some files (*.mbt), weird errors happen, it seems it can not deal with such event
### OS

Operating System: ____

### MoonBit CLI Tools Version

```

```

## Checklist
- [ ] (Optional) My case is minimal enough to be reproducible.

lynzrand:
Closes #1106 

The legacy code cached `ModuleDB` (package database) between different `moon check` runs in watch mode, presumably because it was slow to re-scan the whole project. This database cache is brittle because if a file is added or deleted, it won't be reflected to the cached DB, causing the project to fail in strange ways (like #1106, but should happen before #1076 changed the impl of watching) when a file is deleted and then another is modified.

This PR puts the legacy code path to the more naive way to do this: no caches, rescan everything. It should fix the previous issue at the cost of making watch mode slower. Nevertheless, since these are legacy code, it's expected to be removed soon, so making it slower won't hurt (unless someone complains).

### DIFF
--- a/crates/moonbuild/src/lib.rs
+++ b/crates/moonbuild/src/lib.rs
@@ -39,31 +39,6 @@ pub mod upgrade;
 
 use std::sync::LazyLock;
 
-use sysinfo::{ProcessExt, System, SystemExt};
-
-pub const MOON_PID_NAME: &str = ".moon.pid";
-
-pub fn watcher_is_running(pid_path: &std::path::Path) -> anyhow::Result<bool> {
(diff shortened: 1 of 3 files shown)


lynzrand:
The test error is an error caused by stale changes in `moonbitlang/x` which is out of our control. We will merge this regardless of the test result because the fix is being handled in parallel.

PR #1112, opened 2025-10-21
Simplify issue and PR templates

lynzrand:
Nobody writes them cleanly so we're better off just use simpler templates.

TODO: maybe use the github form variant?


lynzrand:
No changes to code, direclty merging

PR #1167, opened 2025-11-02
feat: honor MOON_HOME and add overrides

fantix:
- Related issues: None 
- PR kind: Feature

## Summary

This PR changes `moon` to practically honor the `MOON_HOME` environment variable for all bundled tool searches. This is a breaking change mostly in a way that `moon` will stop looking for e.g. `moonc` or `moonrun` in `$PATH`, in order to lower the chance to run a tool from a different toolchain. This PR removes most - if not all - of the hard-coded tool names.

Please review each commit separately.

## Design

This PR adds `moonutils::BINARIES` with 2 types of tools:
* `moon_bin`: bundled tools under `$MOON_HOME/bin/`
* `which_bin`: system tools looked up in `$PATH` (with smart extension names on Windows by the `which` crate), optional by field, falling back to the identifier by method (`node()` returns `"node"` if `$PATH` doesn't have Node.js installed).

Always using `BINARIES` gives us the ability to override these tools (on top of `MOON_HOME` override):

* `MOONC_OVERRIDE`
* `MOONCAKE_OVERRIDE`
* `MOONDOC_OVERRIDE`
* `MOONFMT_OVERRIDE`
* `MOONINFO_OVERRIDE`
* `MOONLEX_OVERRIDE`
* `MOONRUN_OVERRIDE`
* `MOONYACC_OVERRIDE`
* `MOON_COVE_REPORT_OVERRIDE`
* `MOON_NODE_OVERRIDE`
* `MOON_PYTHON_OVERRIDE`
* `MOON_GIT_OVERRIDE`

## Breaking changes

* These tools will now get loaded from `$MOON_HOME/bin/` or `~/.moon/bin/`, with `.exe` suffix added on Windows:
   * `moonc`
   * `mooncake`
   * `moondoc`
   * `moonfmt`
   * `mooninfo`
   * `moonrun`
   * `moon_cove_report`
   * UPDATE: before `$MOON_HOME` or `~/.moon`, the CLI will firstly look for such tools in its neighbourhood under the same directory.
* ~If the current `moon` is not the one under `$MOON_HOME/bin/`, an `execvp()` (POSIX-only, subprocess on Windows) will replace the running command with the one from `$MOON_HOME`. See also `reexec_from_moon_home()`.~

## What's Not Fixed

* The current codebase is largely confused with `String` vs `OsStr`, as well as the conversion methods like `PathBuf::display()` vs `PathBuf::to_str()`. This PR did not fix those mistakes, but tends to correct the usages as much as possible whenever such lines are covered.
* With the recent toolchain upgrade, `once_cell` can be replaced with built-in "once" equivalents.

## Metadata

### DIFF
--- a/crates/moon/src/rr_build/dry_run.rs
+++ b/crates/moon/src/rr_build/dry_run.rs
@@ -44,7 +44,12 @@ pub fn print_dry_run<'a>(
         })
         .collect::<Vec<_>>();
 
-    try_debug_dump_build_graph_to_file(graph, &default_files, source_dir);
+    let path_replace_table = moonutil::BINARIES
+        .all_moon_bins()
+        .iter()
+        .map(|(name, path)| (path.to_string_lossy().to_string(), name.to_string()))
+        .collect();
+    try_debug_dump_build_graph_to_file(graph, &default_files, source_dir, path_replace_table);
 
     moonbuild::dry_run::print_build_commands(graph, &default_files, source_dir, target_dir);
 }
--- a/crates/moonbuild-debug/src/graph.rs
+++ b/crates/moonbuild-debug/src/graph.rs
@@ -20,7 +20,6 @@
 
 use std::{
     collections::HashSet,
-    env,
     io::{BufRead, Write},
     path::{Path, PathBuf},
     sync::LazyLock,
@@ -80,22 +79,24 @@ pub fn debug_dump_build_graph(
     graph: &n2::graph::Graph,
     input_files: &[FileId],
     source_dir: &Path,
+    path_replace_table: Vec<(String, String)>,
 ) -> BuildGraphDump {
     let accessible_nodes = dfs_for_accessible_nodes(graph, input_files);
-    generate_from_nodes(graph, accessible_nodes, source_dir)
+    generate_from_nodes(graph, accessible_nodes, source_dir, path_replace_table)
 }
 
 pub fn try_debug_dump_build_graph_to_file(
     build_graph: &n2::graph::Graph,
     default_files: &[n2::graph::FileId],
     source_dir: &Path,
+    path_replace_table: Vec<(String, String)>,
 ) {
     let Some(out_file) = DRY_RUN_TEST_OUTPUT.as_deref() else {
         return;
     };
 
     let file = std::fs::File::create(out_file).expect("Failed to create dry-run dump target");
-    let dump = debug_dump_build_graph(build_graph, default_files, source_dir);
+    let dump = debug_dump_build_graph(build_graph, default_files, source_dir, path_replace_table);
     dump.dump_to(file).expect("Failed to dump to target output");
 }
 
@@ -129,8 +130,9 @@ fn generate_from_nodes(
     graph: &n2::graph::Graph,
     accessible_nodes: impl IntoIterator<Item = BuildId>,
     source_dir: &Path,
+    path_replace_table: Vec<(String, String)>,
 ) -> BuildGraphDump {
-    let normalizer = PathNormalizer::new(source_dir);
+    let normalizer = PathNormalizer::new(source_dir, path_replace_table);
     let mut nodes = vec![];
     for node in accessible_nodes {
         let node = graph.builds.lookup(node).expect("Unknown build in graph");
@@ -174,17 +176,15 @@ fn generate_from_nodes(
 
 struct PathNormalizer {
     canonical: Option<PathBuf>,
-    moon_bin_str: Option<String>,
+    replace_table: Vec<(String, String)>,
 }
 
 impl PathNormalizer {
-    fn new(source_dir: &Path) -> Self {
+    fn new(source_dir: &Path, replace_table: Vec<(String, String)>) -> Self {
         let canonical = dunce::canonicalize(source_dir).ok();
-        let moon_bin = env::current_exe().ok();
-        let moon_bin_str = moon_bin.as_ref().map(|p| p.to_string_lossy().to_string());
         PathNormalizer {
             canonical,
-            moon_bin_str,
+            replace_table,
         }
     }
 
@@ -199,10 +199,9 @@ impl PathNormalizer {
             s = s.replace(prefix_str, ".");
         }
 
-        if let Some(moon) = self.moon_bin_str.as_deref() {
-            s = s.replace(moon, "moon");
+        for (from, to) in &self.replace_table {
+            s = s.replace(from, to);
         }
-
         s = s.replace('\\', "/");
 
         s
--- a/crates/moonbuild/src/dry_run.rs
+++ b/crates/moonbuild/src/dry_run.rs
@@ -56,7 +56,12 @@ pub fn print_build_commands(
         }
     }
 
-    try_debug_dump_build_graph_to_file(graph, default, source_dir);
+    let path_replace_table = moonutil::BINARIES
+        .all_moon_bins()
+        .iter()
+        .map(|(name, path)| (path.to_string_lossy().to_string(), name.to_string()))
+        .collect();
+    try_debug_dump_build_graph_to_file(graph, default, source_dir, path_replace_table);
 }
 
 /// Print run commands from a State


fantix:
@lynzrand is #1184 the preferred way to this one? I'd appreciate some comments from the maintainers rather than silently deprecating community PRs.

lynzrand:
> is #1184 the preferred way to this one?

Nope. That was only implemented for it to work, not for it to last. Your implementation is definitely superior. 

I have done a brief review of the diffs (will do a full review the next workday), here are my comments:

- The overall "concentrate every binary to use in one place" design is solid.

- No problem in the replacement part from my brief review.

- I don't see the point of using a macro to encapsulate the path resolving part of it, when you can definitely do it via functions. e.g., this is fine:

  ```rust
  binaries! { CachedBinaries {
    field_name: definition
  } }

  ==>>

  struct CachedBinaries {
    field_name: PathBuf
  }
  static BINARIES = LazyLock::new(|| CachedBinaries {
    field_name: definition
  });
  ```

  But I don't see the point of writing the `definition`s themselves via macro. You can just write `moon_bin(filename, env)` with a plain function and call it a day. This would be much more readable and maintainable compared to the current implementation.

For why this PR has been "ghosted", it's because we were extremely limited by bandwidth in the recent weeks. There were just too many things to implement before we had the time to look at community contributions.

fantix:
Roger; that's totally understandable, thank you for the update! I'll rebase this PR onto latest main today and fix accordingly for your review.

crates/moon/src/cli/coverage.rs
```diff
@@ -162,7 +162,7 @@ fn run_coverage_report_command(
     cwd: &Path,
     dry_run: bool,
 ) -> std::io::Result<std::process::ExitStatus> {
-    let mut cmd = std::process::Command::new("moon_cove_report");
+    let mut cmd = std::process::Command::new(moonutil::BINARIES.moon_cove_report.as_os_str());
```
lynzrand: nit: `PathBuf` is already `AsRef<OsStr>`, so in theory this conversion is not needed (You can write `&BINARIES.moon_cove_report` and so on)

crates/moon/src/main.rs
```diff
@@ -83,9 +83,56 @@ fn init_tracing() -> Box<dyn Any> {
...
+            }
+        }
+    }
+}
+
 pub fn main() {
     panic::setup_panic_hook();
 
+    // Check MOON_HOME override and re-exec if needed
+    reexec_from_moon_home();
```
lynzrand: I personally don't think integrating this logic in `moon` is a good idea. A `rustup`-style executable multiplexer would be better.

I once had written one myself, but that project has been long-abandoned due to various changes in MoonBit's release methods. Feel free to make use of it: https://github.com/lynzrand/lunik

crates/moonutil/src/binaries.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    pub fn git_or_default(&self) -> PathBuf {
+        self.git
+            .clone()
+            .unwrap_or_else(|| get_fallback_binary("git"))
+    }
+}
+
+pub static BINARIES: CachedBinaries = CachedBinaries {
+    moonbuild: LazyLock::new(|| moon_bin("moon", "MOON_OVERRIDE")),
```
lynzrand, line 100: I guess that `moon` self-call should prefer `current_exe` first before others?

crates/moonutil/src/scan.rs
```diff
@@ -506,6 +505,22 @@ fn scan_one_package(
...
+                        format!(
+                            "cannot decode {} path: {:?}",
+                            stringify!($field),
+                            crate::BINARIES.$field
+                        )
+                    })
+            }
+        }};
+    }
+    let moonrun = stringify_bin!(moonrun);
```
lynzrand, line 532: nit: I was about to complain about random macros occuring, but since this is legacy code I'll just let it go

lynzrand (changes requested):

crates/moon/src/main.rs
fantix: Yeah, I also agree that execvp() could be surprising to many. Now dropped.

crates/moonutil/src/binaries.rs
fantix: This one is kinda connected to the execvp() thing above - now that we dropped the re-exec, I added a sibling lookup instead: https://github.com/moonbitlang/moon/pull/1167/commits/d1c6ae8da904f1f2dbed0179552eee3f4add88d8

crates/moonutil/src/scan.rs
fantix: Was it? Took me a while to get it right 😏 anyways, I think in the end of the day, all path strings should go away; instead, `OsStr`-compatible references won't need this macro.

crates/moon/src/main.rs
chawyehsu: It might also break moonup integrating this reexec in `moon`. To some extent, [`moonup`](https://github.com/chawyehsu/moonup) is trying to be the `rustup` for MoonBit, acknowledging prior work including [`multimoon`](https://github.com/lone-outpost-oss/multimoon), etc,.

lynzrand: > It might also break moonup integrating this reexec in `moon`. To some extent, [`moonup`](https://github.com/chawyehsu/moonup?rgh-link-date=2025-11-11T02%3A09%3A18Z) is trying to be the `rustup` for MoonBit, acknowledging prior work including [`multimoon`](https://github.com/lone-outpost-oss/multimoon?rgh-link-date=2025-11-11T02%3A09%3A18Z), etc,.

If `moonup` is actively maintained, you might want to reuse the multiplexer code in `lunik` so that it multiplexes executables like `rustup`. This way, (accompanied by this PR and friends) you can write `moon +nightly test` and similar.

crates/moon/src/cli/coverage.rs
```diff
@@ -162,7 +162,7 @@ fn run_coverage_report_command(
     cwd: &Path,
     dry_run: bool,
 ) -> std::io::Result<std::process::ExitStatus> {
-    let mut cmd = std::process::Command::new("moon_cove_report");
+    let mut cmd = std::process::Command::new(&*moonutil::BINARIES.moon_cove_report);
```
lynzrand, line 165: Nit: `&BINARIES.moon_cove_report` should be enough, Rust has auto-deref. No need too change as this runs just as fine.

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
```diff
@@ -799,12 +799,16 @@ impl<'a> BuildPlanLowerContext<'a> {
 
         // Note: at this point, we have our TCC command.
         // However, this command should be executed when the user runs the final
-        // executable, not in this build graph. Thus, we need to put them into
+        // executable, not in this build graph. Thus, we need to put them int&self.binarieso
```
lynzrand: Whoops? what's going on here

crates/moonutil/src/scan.rs
lynzrand: Yeah, you can probably feel that this part writes *different* and you have encountered duplicated logic. There has been a rewrite going on for months, and we're nearing the end of it.

crates/moon/src/main.rs
chawyehsu: > This way, (accompanied by this PR and friends) you can write moon +nightly test and similar.

`moonup` already supports channels and the [`+<toolchain_spec>` syntax](https://github.com/chawyehsu/moonup?tab=readme-ov-file#run-a-command-with-a-specific-toolchain), will visit `lunik` to see if anything `moonup` can be improved.

We encountered a problem on the `moonup` side from the very beginning that was difficult to solve, _the isolation of the toolchain_[^1][^2]. I’ve always hoped the team could make improvements in this area. Although it may not happen in the short term, it’s still something worth looking forward to.

[^1]: https://github.com/moonbitlang/moon/discussions/327
[^2]: https://github.com/chawyehsu/moonup/issues/7

lynzrand: Oh! It was too small for my last skim through the docs. Then there's probably little in Lunik for you to copy, you've got everything. Maybe the multiplexer is worth a more significant place in the docs (like, `cargo` and `rustc` are actually multiplexing shims created by `rustup` that calls the real underlying binary), so users will know `moonup` is not a simple version manager but also a muxer.

crates/moon/src/cli/coverage.rs
fantix: I was expecting that to work too, but actually:

```
error[E0277]: the trait bound `LazyLock<PathBuf>: AsRef<std::ffi::OsStr>` is not satisfied
   --> crates/moon/src/cli/coverage.rs:165:47
    |
165 |     let mut cmd = std::process::Command::new(&moonutil::BINARIES.moon_cove_report);
    |                   --------------------------  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `AsRef<std::ffi::OsStr>` is not implemented for `LazyLock<PathBuf>`
    |                   |
    |                   required by a bound introduced by this call
    |
    = note: required for `&LazyLock<PathBuf>` to implement `AsRef<std::ffi::OsStr>`
note: required by a bound in `std::process::Command::new`
   --> /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/process.rs:655:5
help: consider dereferencing here
    |
165 |     let mut cmd = std::process::Command::new(&*moonutil::BINARIES.moon_cove_report);
    |                                               +

For more information about this error, try `rustc --explain E0277`.
error: could not compile `moon` (bin "moon") due to 1 previous error
```

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
fantix: Gotta be a mousefart 🤦‍♂️

fantix:
Fixed and rebased, thank you @lynzrand for the prompt review! Looks like the stable-test failure is irrelevant; though I do need to fix the failure on Windows.

crates/moon/src/cli/coverage.rs
lynzrand: Well then ignore me and trust `rustc` for this ;)

fantix:
Rebased again and added 3 commits.

lynzrand:
This is very helpful! We will review, check and merge this shortly after

fantix:
Thank you, @lynzrand! I was looking into the rr-parity failure, but now that it's in `main`, maybe your team wanna take it from there.

lynzrand:
Yes, we're doing bug bashing in `rr-parity` (i.e. parity test for `NEW_MOON=1` tests). Thanks for your contribution!

PR #1175, opened 2025-11-05
fix: not use global gitignore when copy files

Yu-zh:
## Summary

not use global gitignore when copy files during `cargo test`


lynzrand (approved):

lynzrand:
uh-oh. we're using manual semi-linear merging, so merge commits are rejected XD

PR #1182, opened 2025-11-06
fix(test): handle `#skip` attribute in gen-test-driver

Yu-zh:
## Summary

`moonc` will not filter skipped test, so `moon` needs to do that.

## Metadata


lynzrand:
Will the `--include-skipped` flag be handled in another PR?

Yu-zh:
> Will the `--include-skipped` flag be handled in another PR?

yep. this just fixes the CI.

PR #1185, opened 2025-11-07
feat: --include-skipped for `moon test`

Yu-zh:
- Related issues: None
- PR kind: feature

## Summary

Implemented `--include-skipped` for `moon test`. The handling `#skip` attribute implemented #1182 is not appropriate because it filters the tests when generating the metadata and doesn't fix the test filter. Instead this PR use the test filter to handle the `#skip` and `--include-skipped` information properly.

## Metadata

### DIFF
--- a/docs/manual-zh/src/commands.md
+++ b/docs/manual-zh/src/commands.md
@@ -253,8 +253,8 @@ Test the current package
 
 * `-p`, `--package <PACKAGE>` — Run test in the specified package
 * `-f`, `--file <FILE>` — Run test in the specified file. Only valid when `--package` is also specified
-* `-i`, `--index <INDEX>` — Run only the index-th test in the file. Only valid when `--file` is also specified
-* `--doc-index <DOC_INDEX>` — Run only the index-th doc test in the file. Only valid when `--file` is also specified
+* `-i`, `--index <INDEX>` — Run only the index-th test in the file. Only valid when `--file` is also specified. Implies `--include-skipped`
+* `--doc-index <DOC_INDEX>` — Run only the index-th doc test in the file. Only valid when `--file` is also specified. Implies `--include-skipped`
 * `-u`, `--update` — Update the test snapshot
 * `-l`, `--limit <LIMIT>` — Limit of expect test update passes to run, in order to avoid infinite loops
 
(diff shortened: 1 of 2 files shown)


crates/moon/tests/test_cases/test_filter/include_skip_test.rs
```diff
@@ -0,0 +1,23 @@
+use crate::{TestDir, get_stdout};
+
+#[test]
+fn test_skip_included() {
+    let dir = TestDir::new("test_filter/skip_test");
+    let stdout = get_stdout(&dir, ["test", "--verbose", "--include-skipped"]);
+    assert!(
+        stdout.contains("hello_0"),
+        "Expected hello_0 to be skipped, got:\n{stdout}j"
```
lynzrand: Message forgot to change

crates/moon/tests/test_cases/test_filter/include_skip_test.rs
```diff
@@ -0,0 +1,23 @@
...
+    );
+    assert!(
+        stdout.contains("hello_1"),
+        "Expected hello_1 to run, got:\n{stdout}"
+    );
+    assert!(
+        stdout.contains("hello_2"),
+        "Expected hello_2 to be skipped, got:\n{stdout}"
```
lynzrand: This too

crates/moonbuild/src/entry.rs
```diff
@@ -813,8 +814,12 @@ pub fn run_test(
                     let ranges = vec![filter_index..(filter_index + 1)];
                     test_args.file_and_index.push((file_name.clone(), ranges));
```
lynzrand, line 816: Didn't check `has_skip` here. Behavior different from new one.

Yu-zh: I have changed the behavior in latest commit - when a particular testcase is specified using `-f file -i index`, then it ignores the `#skip` attribute. Is there a place where I need to document this behavior?

lynzrand: Hmm. I don't think we have a documentation of `moon test` behavior yet. 

Does it fit to add something like a `warn!("No tests are selected, the filter specified tests but were skipped due to '#skip', you might want to use '--include-skipped' when testing")` here?

lynzrand: NB: Will let `--index`/`--doc-index` imply `--include-skipped`

PR #1188, opened 2025-11-09
fix: block SIGCHLD only on macOS

tonyfettes:
- Related issues:
  - https://github.com/moonbitlang/maria/issues/255
  - https://github.com/moonbitlang/moon/pull/1156
- PR kind: Bugfix

## Summary

We should only block sigchild on macOS - blocking it on Linux will cause moon test hangs when the test spawns child processes.

## Metadata


lynzrand:
aw shit, you're probably right.

PR #1192, opened 2025-11-10
[RR] Fix `moon bundle` behavior

lynzrand:
- Related issues: None 
- PR kind: Bugfix 

## Summary

The behavior of `moon bundle` is corrected to match the legacy one. Main things fixed:

- Fixed output directory of bundled `.core` from `target/<backend>/*.core` to `target/<backend>/<opt>/<action>/*.core`
- Fixed the action input to `moon bundle` from `Build` to `Bundle`.
- Fixed `bundle` to skip virtual packages (`moonbitlang/core/abort`) when bundling.

This should make installing the tool chain working with the new backned.

## Metadata


lynzrand:
Checked the behavior to be identical to legacy.

PR #1200, opened 2025-11-12
CD: remove deprecated macos-13 in cd.yml

illusory0x0:


lynzrand:
Thanks, that helps very much!

PR #1211, opened 2025-11-13
fix(rr): not check on third party blackbox test when `moon check`

Yu-zh:
- Related issues: None
- PR kind: Bugfix

## Summary

Currently when executing `moon check` in a module with third party dependency, it will also run `moonc check` on the third party's blackbox/whitebox tests, which leads to a test failure in `test_cases::test_specify_source_dir_with_deps`

## Metadata


lynzrand:
yeah i missed this on my last fix. thanks ;)

PR #1237, opened 2025-11-19
test(rr): add whitelist in failure test

Yu-zh:
- Related issues: None

## Summary

Added a whitelist for testcases that RR failed. These failures are not fatal so at the end when we switch to use RR by default, we can just promote these tests.

## Metadata


lynzrand:
maybe another file so editing is easier?

Yu-zh:
> maybe another file so editing is easier?

Done

xtask/src/test_rr_whitelist.rs
lynzrand, line 1: I actually might lean to using `include_str!` (https://doc.rust-lang.org/std/macro.include_str.html) for this, if you don't mind ;)

Yu-zh: Make sense. Just fixed.

lynzrand: lgtm

PR #1244, opened 2025-11-20
Fix package linkability detection

lynzrand:
- Related issues: None 
- PR kind: Bugfix 

## Summary

This PR fixes linking-related issues for RR, and passes `test_native_backend_*` (within error margins, wontfix).

- Fix linkable package condition to checking `pkg.link`'s specific fields against backend
- Fix `moon test --dry-run` to still print in tcc mode regardlessly
- Deduplicate C stubs within the same package (mainly bb/wb test vs. source)
- Fix shared C stub compilation input

Also:

- Fixed the logic regarding bin dependency target resolving `moon build-bin-dep`
- Cleaned up `calculate_user_intent` callback args.

## Things needing second thought

Might need a better way to filter the input nodes when loweing build plan to N2 build graph.

## Metadata


lynzrand:
seems like this broke other tests. looking into it.

lynzrand:
Yeah, I think this should fix it

Yu-zh (approved):

PR #1245, opened 2025-11-20
fix problem with `--`

Yu-zh:
- Related issues: None

## Summary

Fixed: `test_cases::target_backend::test_target_backend`

With this PR, when using `node` for executing JS file, it will not emit `--`. When using `moonrun` for executing wasm file, it will only emit `--` when there's argv to pass.

## Metadata

### DIFF
--- a/crates/moon/tests/test_cases/target_backend/mod.rs
+++ b/crates/moon/tests/test_cases/target_backend/mod.rs
@@ -68,7 +68,7 @@ fn test_target_backend() {
             moonc build-package ./lib/hello.mbt -o ./target/js/release/build/lib/lib.core -pkg hello/lib -pkg-sources hello/lib:./lib -target js -workspace-path .
             moonc build-package ./main/main.mbt -o ./target/js/release/build/main/main.core -pkg hello/main -is-main -i ./target/js/release/build/lib/lib.mi:lib -pkg-sources hello/main:./main -target js -workspace-path .
             moonc link-core ./target/js/release/build/lib/lib.core ./target/js/release/build/main/main.core -main hello/main -o ./target/js/release/build/main/main.js -pkg-config-path ./main/moon.pkg.json -pkg-sources hello/lib:./lib -pkg-sources hello/main:./main -target js
-            node ./target/js/release/build/main/main.js --
+            node ./target/js/release/build/main/main.js
         "#]],
     );
 }


crates/moon/src/run/runtime.rs
lynzrand, line 47: The original intention to add the `--` was to make the consumers of the function not needing to bother checking, and just directly append their args to this command. So, maybe we can just modify the legacy one instead?

crates/moon/src/cli/run.rs
```diff
@@ -474,7 +474,17 @@ fn get_run_cmd(
...
+        // Add separator for Wasm/WasmGC backends to separate moonrun args from program args
+        if matches!(
+            build_meta.target_backend,
+            moonbuild_rupes_recta::model::RunBackend::Wasm
+                | moonbuild_rupes_recta::model::RunBackend::WasmGC
+        ) {
+            cmd.command.arg("--");
+        }
+        cmd.command.args(argv);
+    }
```
lynzrand: -> see https://github.com/moonbitlang/moon/pull/1245/files#r2548414217, i would like to avoid the special case checking here

crates/moonbuild/src/dry_run.rs
```diff
@@ -98,15 +98,26 @@ pub fn print_run_commands(
...
+            } else if args.is_empty() {
                 watfile
+            } else {
+                format!("{watfile} {}", args.join(" "))
             };
-            if !args.is_empty() {
-                moonrun_command = format!("{moonrun_command} {}", args.join(" "));
-            }
 
             println!("{moonrun_command}");
```
lynzrand: Ideally it might be preferred to just join the args in a vec and use `shlex` for it? Don't know if it's worth the trouble

crates/moon/src/cli/run.rs
Yu-zh: so you mean keeping it as `moonrun xxx.wat --`?

lynzrand: yes, at least that's what I intended to do. feel free to have different ideas ;)

Yu-zh: I'm fine either way. I will just deal with `node` then.

lynzrand (approved):
lgtm

PR #1258, opened 2025-11-27
[RR] Implement single-file mode

lynzrand:
- Related issues: None 
- PR kind: feature 

## Summary

This PR implements `moon {check,test,run}`'s single file mode in RR.

To implement single file mode, I added:
- a separate single-file package relationship resolver (because single-file modes don't have a regular project structure)
- markers for single-file modules/packages

Everything beyond package resolving is the exact same pipeline as non-single-file mode.

Tests were edited to accommodate RR's new behaviors. 3 tests were marked as won't fix, all of which are caused by project layout changes in `moon run`. `moon check` and `moon test` are compatible without minor edits of test criteria.

- test_cases::test_moon_run_single_file_dry_run
- test_cases::value_tracing::test_tracing_value_for_single_file
- test_cases::value_tracing::test_tracing_value_for_single_file_dry_run

## Metadata

### DIFF
--- a/xtask/rr_wont_fix.txt
+++ b/xtask/rr_wont_fix.txt
@@ -33,3 +33,8 @@ test_cases::targets::test_many_targets
 # test order different
 test_cases::test_virtual_pkg
 
+# `moon run` single-file mode is too different from others
+# dry-run output and artifact layout is different
+test_cases::test_moon_run_single_file_dry_run
+test_cases::value_tracing::test_tracing_value_for_single_file
+test_cases::value_tracing::test_tracing_value_for_single_file_dry_run


lynzrand:
This PR marks the end of #802.

Yu-zh:
Running single file seems changed from using tcc to using cc. But I guess that's fine because it's single file and should not be super slow like testing.

Yu-zh (approved):

lynzrand:
> Running single file seems changed from using tcc to using cc. But I guess that's fine because it's single file and should not be super slow like testing.

this is a bug. allow me to verify and check it.

lynzrand:
`run` at that time defaulted to release mode. switched to debug mode and i think we're good to go.

bobzhang:
> Running single file seems changed from using tcc to using cc. But I guess that's fine because it's single file and should not be super slow like testing.

@Yu-zh My intuition is that even for a single file, cc instead of tcc could still cause significant delay

PR #1263, opened 2025-11-28
switch to RR backend by default

Yu-zh:
## Summary

This PR switched to RR backend by default. The legacy moon can be enabled by using `NEW_MOON=0`. The snapshot diffs seem to be more than expected.

### DIFF
--- a/crates/moon/tests/test_cases/fuzzy_matching.in/moon.test
+++ b/crates/moon/tests/test_cases/fuzzy_matching.in/moon.test
@@ -37,7 +37,7 @@
   $ moon test -p asdf --sort-input --no-parallelize
   Total tests: 0, passed: 0, failed: 0.
   
-  [33m WARN[0m no package found matching the given filters
+  [33m WARN[0m package `asdf` not found, make sure you have spelled it correctly, e.g. `moonbitlang/core/hashmap`(exact match) or `hashmap`(fuzzy match)
   Warning: no test entry found.
   
   $ moon test -p y -f asdf --sort-input --no-parallelize
--- a/crates/moon/src/cli/test.rs
+++ b/crates/moon/src/cli/test.rs
@@ -870,7 +870,10 @@ fn apply_list_of_filters(
         }
     } else {
         // No package matched
-        warn!("no package found matching the given filters");
+        warn!(
+            "package `{}` not found, make sure you have spelled it correctly, e.g. `moonbitlang/core/hashmap`(exact match) or `hashmap`(fuzzy match)",
+            package_filter.join(", ")
+        );
     }
     trace!("finished building package directive");
 


crates/moon/tests/test_cases/cond_comp.in/moon.test
```diff
@@ -14,12 +14,12 @@
   moonc link-core ./target/wasm-gc/release/build/lib/lib.core ./target/wasm-gc/release/build/main/main.core -main username/hello/main -o ./target/wasm-gc/release/build/main/main.wasm -pkg-config-path ./src/main/moon.pkg.json -pkg-sources username/hello/lib:./src/lib -pkg-sources username/hello/main:./src/main -target wasm-gc
   
   $ moon build --dry-run --target js --sort-input --nostd
-  moonc build-package ./src/lib/all.mbt ./src/lib/js_and_release.mbt ./src/lib/js_or_wasm.mbt ./src/lib/only_js.mbt ./src/lib/wasm_release_or_js_debug.mbt -o ./target/js/release/build/lib/lib.core -pkg username/hello/lib -pkg-sources username/hello/lib:./src/lib -target js -workspace-path .
```
lynzrand, line 17: I saw this in wonfix list. So this is because legacy implements conditional compilation wrong, is that true?

crates/moon/tests/test_cases/fuzzy_matching.in/moon.test
```diff
@@ -35,13 +35,8 @@
...
   $ moon test -p asdf --sort-input --no-parallelize
-  [255]
-  
-  error: package `asdf` not found, make sure you have spelled it correctly, e.g. `moonbitlang/core/hashmap`(exact match) or `hashmap`(fuzzy match)
-  
-  $ moon test -p y -f asdf --sort-input --no-parallelize
   Total tests: 0, passed: 0, failed: 0.
   
-  Warning: cannot find file `asdf` in package `username/hello/x/y`, `username/hello/y`, --file only support exact matching
+  [33m WARN[0m no package found matching the given filters
```
lynzrand: Do we need to update the message? This sounds like a regression

crates/moon/tests/test_cases/fuzzy_matching/mod.rs
```diff
@@ -19,3 +20,23 @@ fn test_fuzzy_matching() {
...
+            "-f",
+            "asdf",
+            "--sort-input",
+            "--no-parallelize",
+        ],
+    );
+    assert!(output.contains("error: Cannot filter by file or index when multiple packages are specified. Matched packages:"));
+    assert!(output.contains("username/hello/x/y"));
+    assert!(output.contains("username/hello/y"));
+}
```
lynzrand: This was changed to this because the order was inconsistent because it was using HashMap. Pushed BTreeMap to it, which should fix.

lynzrand:
If you consider it stable enough for everyday use, you can merge it. It's your project now ;)

crates/moon/tests/test_cases/fuzzy_matching.in/moon.test
Yu-zh: I will improve the warning message.

PR #1270, opened 2025-12-01
Simplify & fix warn-list

lynzrand:
- Related issues: None 
- PR kind: Bugfix 

## Summary

`moonc` has merged alert-list to warn-list, and `@*` only sets warninngs **that are currently open** to error. Thus, we can simplify the warn list behavior a lot.

Also: `test_import_all` is renamed, so this fails in main branch.

## Metadata


Yoorkin (approved):
LGTM

lynzrand:
Merging because CI is failing because of another ICE, and it was tested to be correct on yesterday's nightly.

PR #1278, opened 2025-12-02
chore: use defensive namespace

peter-jerry-ye:
- Related issues: None 
- PR kind: 

## Metadata


lynzrand (approved):
LGTM.

NB: In `builtin`, `@moonbitlang/core/builtin.` will be replaced with an empty string, so this works in `builtin`.

PR #1279, opened 2025-12-03
[RR] Skip pre-build script output when formatting

### LINKED ISSUE #1275: Avoid stale-loop between pre-build file generation and moon fmt
## Actual behavior

In projects where pre-build steps generate source files that are also formatted by `moon fmt`, the two actions can conflict and cause each other to always see files as stale:

- The pre-build step generates files.
- `moon fmt` reformats those generated files.
- On the next run, the pre-build step sees the formatted files as stale compared to its own expected output and regenerates them.
- `moon fmt` then sees the regenerated files as stale again and reformats them.

As a result, both the generator and the formatter keep re-running instead of converging to a stable, up-to-date state.

## Expected Behavior

- Pre-build generated files and `moon fmt` should not continuously invalidate each other.
- The build graph should converge so that:
  - After the first successful run, subsequent runs do not keep re-running both the generator and `moon fmt` for the same files, or
  - Pre-build generated files that are not intended to be formatted are automatically skipped by `moon fmt`.

## Steps to Reproduce

1. Use a project where a pre-build step generates files that are also picked up by `moon fmt`.
2. Run `moon fmt` (or a command that triggers both generation and formatting).
3. Run the same command again.
4. Observe that both generation and formatting keep running because each sees the other's output as stale.

## Environment

OS: Any

Related parent issue context: generated pre-build files should be skipped by `moon fmt` so that they do not fight with the generator.

lynzrand:
- Related issues: #1275 
- PR kind: Feeature 

## Summary

Closes #1275 

This PR skips prebuild output files when formatting. This should remove the stale loop between formatting and prebuild outputting overwritting each other.

## Metadata

### DIFF
--- a/crates/moonrun/tests/test.rs
+++ b/crates/moonrun/tests/test.rs
@@ -260,3 +260,24 @@ fn test_moon_run_with_is_windows() {
     };
     assert!(actual_result == expected_result);
 }
+
+#[test]
+fn test_moon_fmt_skips_prebuild_output() {
+    // Prepare a temp copy of the test case
+    let dir = TestDir::new("test_fmt_skip_prebuild_output");
+
+    // The prebuild command is a NOOP; we intentionally wrote a sloppy file as the "generated" output.
+    // Ensure the source remains sloppy after fmt (formatter must skip prebuild outputs).
+    let generated_src = dir.join("main/generated.mbt");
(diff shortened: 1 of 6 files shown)


Yu-zh (approved):

Yu-zh:
seems legacy `moon fmt` also has this problem so the CI doesn't pass.

lynzrand:
Oops. Let me fix that.

PR #1280, opened 2025-12-03
fix(rr): not emit -o .mi when --no-mi is passed

Yu-zh:
- Related issues: None
- PR kind: Bugfix

## Summary

The `moonc check` command is passed `--no-mi` when checking a virtual package's implementation. However, `-o .mi` is also passed which overrides `--no-mi` and might create a mi file that overwrites the virtual package's `mi`. This PR fixes this bug by making sure that when `--no-mi` is enabled, `-o .mi` will not be passed.

The fix is bit hacky though. Because there is no need to calculate the artifact `.mi` path when no `.mi` is generated. I will clean this up later.

## Metadata

### DIFF
--- a/xtask/rr_expected_failures.txt
+++ b/xtask/rr_expected_failures.txt
@@ -16,6 +16,3 @@ test_cases::dummy_core::test_dummy_core
 
 # moonrun binary improvements
 test_cases::target_backend::test_target_backend
-
-# macos failure that cannot reproduce locally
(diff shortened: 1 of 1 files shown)


lynzrand (approved):
LGTM. 

NB: Maybe in the future we should review how "dumb" should the compiler command wrappers should be, but it's okay for this quickfix.

PR #1281, opened 2025-12-04
support for indirect dependency

Yu-zh:
- Related issues: None
- PR kind: feature

## Summary

This PR makes the build system generate `all_pkgs.json` before executing `moonc <command>`. The information in the generated `all_pkgs.json` is a subset of `packages.json`, and the implementation mostly follows the generation for `all_pkgs.json`.

## Metadata

### DIFF
--- a/crates/moonbuild-rupes-recta/src/all_pkgs.rs
+++ b/crates/moonbuild-rupes-recta/src/all_pkgs.rs
@@ -16,6 +16,17 @@
 //
 // For inquiries, you can contact us via e-mail at jichuruanjian@idea.edu.cn.
 
+//! Generation of `all_pkgs.json` metadata file.
+//!
+//! This module provides functionality to generate the `all_pkgs.json` file,
+//! which is a metadata file used by `moonc` to resolve indirect dependencies.
+//! The file contains information about all packages in the build, including
+//! their root modules, relative paths, and artifact locations (`.mi` files).
+//!
+//! The `all_pkgs.json` file is generated in the target directory during the
+//! build process and is consumed by the MoonBit compiler for dependency
+//! analysis.
+
 use moonutil::common::TargetBackend;
 use serde::{Deserialize, Serialize};
 


lynzrand:
Just wondering -- if we need to pass this to `moonc` on every build command, why not just put it in `rr_build::execute_build` or somewhere like there, so you only add once place to edit?

---

edit: Theoretically we can put `packages.json` unconditionally there too...?

Yu-zh:
> why not just put it in `rr_build::execute_build` or somewhere like there

There are special case like `moon fmt` that doesn't need it. And even if I put it in `rr_build::execute_build`, there are still a few places missing.

lynzrand:
Hmmm. 

I originally thought that `fmt` is small enough as an exception to ignore, as it doesn't do any bad, but in second thought I remembdered that it doesn't even have a `BuildMeta` IIRC. Then I think this behavior is correct. 

We need to say somewhere in the docs it's needed for every `moonc` command, though.

Yu-zh:
> We need to say somewhere in the docs it's needed for every `moonc` command, though.

Thanks for this reminder. I just added `indirect-deps.md` which documents about it with the help of AI.

crates/moonbuild-rupes-recta/src/all_pkgs.rs
```diff
@@ -0,0 +1,80 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+// For inquiries, you can contact us via e-mail at jichuruanjian@idea.edu.cn.
+
```
lynzrand, line 18: Add a module doc comment here, or maybe split `metadata.rs` into two and put the contents there?

PR #1283, opened 2025-12-04
Move MSVC compilation temp dir to per-package

lynzrand:
- Related issues: None 
- PR kind: Bugfix

## Summary

The issue was spotted by @peter-jerry-ye in `mooncakes-dashboard`.

RR accidentally used the same temporary directory across pacakges when compiling for MSVC. MSVC will use the temp dir to store intermediate files of the compilation, including a temporary object file if compiling *and* linking C files within the same compiler call. If two C files with the same name are compiled at the same time, and they're using the same temp dir, the output filename of them will be the same. As a result, one MSVC process will accidentally overwrite or delete the output file of another MSVC process.

This PR corrects this behavior by forcing each MSVC invocation to use their own package directory as temp dir. This way, different MSVC invocations using C input with the same name will not conflict with each other.

TODO: How can we test this? 

## Metadata


lynzrand:
Added a best-effort test that triggers like 60 MSVCs at once.

PR #1298, opened 2025-12-09
fix(rr): Add moon.pkg.json as dependency of some build commands

lynzrand:
- Related issues: None 
- PR kind: Bugfix 

## Summary

Although any changes in the build graph would already be captured by the compiler, we add dependency for `moonc` build commands to the package config, just in case that `moonc` reads it during compilation.

## Metadata


lynzrand:
Didn't have time on this, was busy on docs. Legacy didn't have this feature, maybe we need to promote tests and mark it as wontfix for legacy?

Yu-zh:
> Didn't have time on this, was busy on docs. Legacy didn't have this feature, maybe we need to promote tests and mark it as wontfix for legacy?

No worries. I will take care of it.

Issue #954, opened 2025-08-07
`test failure` is expected to report a precise location


bobzhang:
# Feature Request

```
test {
  fail("Test failed")
}
```
failed log:
```
test moonbitlang/core/builtin/failure.mbt::0 failed: /Users/hongbozhang/git/core/builtin/failure.mbt:56:3-56:22 
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^<---- nice to have a location, so we can navigate to the test block quickly 
FAILED: Test failed
Total tests: 1, passed: 0, failed: 1.
```

## Checklist

- [x] I have added all necessary details to make it easy for the maintainers to understand my request.

lynzrand:
https://github.com/moonbitlang/moon/blob/22d62c27cd640e4293eb935d4def9f70c0539241/crates/moonbuild/src/entry.rs#L1026-L1039

The thing printed here is the file and the **index** of the test block (because we don't mandate a name for each test block), so technically this gives enough info about which test block failed, but practically it's hard to navigate.

Currently, test metadata received from the compiler **does not** contain the position of the tests. To solve this problem, we need to:

- Add position metadata to `moonc gen-test-info`
- Integrate the newly-generated metadata in test result printing (currently it does not read about the `moonc` metadata, and the related types are rigid and passed across multiple layers)

Due to the current status of test result printing code, I suggest we postpone it until the refactor at #802 finishes, and we have a cleaner test result printing routine to work with.

bobzhang:
given filename is already there, we only need add a line number here, this should be a quick fix


```
#filename:lineno index
lisp_interpreter/sexp.mbt:340 7
                                            
^^^^^^^^^^^^^^^^^^<------file name 
```

vscode terminal is pretty eays to pick it up

cc @Yoorkin

lynzrand:
I am unsure if it is really a *quick* fix, but I can try implementing it since this is now marked as high prio.

lynzrand:
FTR: Will perform the following changes:

- Add position (line number) metadata in `moonc`
- Pass the metadata to the place where we print test result
- Edit the printing function: Separate the package name and file name (to reduce ambiguity), add line number info

PR #1301, opened 2025-12-10
fix row/column number off-by-one error for new relative source location

Guest0x0:
In #1296, the location obtained from new, relative path based format is one-based, while the internal representation is zero-based. This problem is causing Nightly CI failure. This PR fixes the issue, turning one-based row/column number into zero based when resolving source location.

### DIFF
--- a/crates/moon/tests/test_cases/run_doc_test/mod.rs
+++ b/crates/moon/tests/test_cases/run_doc_test/mod.rs
@@ -28,7 +28,7 @@ fn test_run_doc_test() {
             +1256
             ----
 
-            [username/hello] test lib/hello.mbt:19 (#2) failed: $ROOT/src/lib/hello.mbt:22:5-22:30 FAILED: this is a failure
+            [username/hello] test lib/hello.mbt:19 (#2) failed: @username/hello/lib_blackbox_test:hello.mbt:22:5-22:30 FAILED: this is a failure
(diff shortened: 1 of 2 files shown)


crates/moon/tests/test_cases/run_doc_test/mod.rs
```diff
@@ -28,15 +28,15 @@ fn test_run_doc_test() {
             +1256
             ----
 
-            [username/hello] test lib/hello.mbt:19 (#2) failed: $ROOT/src/lib/hello.mbt:22:5-22:30 FAILED: this is a failure
+            [username/hello] test lib/hello.mbt:19 (#2) failed: @username/hello/lib_blackbox_test:hello.mbt:22:5-22:30 FAILED: this is a failure
```
lynzrand, line 31: Is this diff caused by the new relative-path-based location?

The original result format is the result of this issue: https://github.com/moonbitlang/moon/issues/954. Do you want to still output the absolute or manifest-relative path of this file, or is the current filename-only format good enough?

lynzrand (approved):

crates/moon/tests/test_cases/run_doc_test/mod.rs
Guest0x0: This is intended, and one of the goals of migrating to relative path in source location. One direct benefit is: this kind of test snapshot with source location is now stable across different machines

PR #1303, opened 2025-12-10
fix: building plan mishandled virtual being implemented

Yu-zh:
- Related issues: None
- PR kind: Bugfix

## Summary

When build an implementation package with a third-party's virtual, the virtual package isn't added to the build plan. This PR fixes this bug.

## Metadata

### DIFF
--- a/crates/moonbuild-rupes-recta/src/discover/model.rs
+++ b/crates/moonbuild-rupes-recta/src/discover/model.rs
@@ -102,10 +102,6 @@ impl DiscoveredPackage {
         self.raw.implement.is_some()
     }
 
-    pub fn get_virtual_impl(&self) -> Option<&String> {
-        self.raw.implement.as_ref()
-    }
-
     /// Get whether if the package has a concrete implementation, i.e. moonbit
     /// code to compile.
     ///


crates/moonbuild-rupes-recta/src/build_plan/builders.rs
```diff
@@ -189,6 +189,19 @@ impl<'a> BuildPlanConstructor<'a> {
             let dep_node = self.need_node(BuildPlanNode::BuildVirtual(target.package));
             self.add_edge(node, dep_node);
         }
+
+        // If the given target implements a virtual package, we need to build
+        // the virtual package's interface first.
+        if let Some(virtual_pkg) = pkg.get_virtual_impl() {
```
lynzrand: IIRC we have a resolved package ID for it, at `pkg_rel.virtual_impl[pkg_id]`, so no need to query the package name string here?

Actually, all stringly-typed package references should have been already solved in `pkg_solve`. If it's not already available here, we should modify `pkg_solve` to solve this for us.

lynzrand (changes requested):

crates/moonbuild-rupes-recta/src/build_plan/builders.rs
```diff
@@ -241,6 +249,18 @@ impl<'a> BuildPlanConstructor<'a> {
...
+            let vpkg_id = self
+                .input
+                .pkg_dirs
+                .get_package_id_by_name(virtual_pkg)
+                .expect("Virtual package should exist");
+            let dep_node = self.need_node(BuildPlanNode::BuildVirtual(vpkg_id));
+            self.add_edge(node, dep_node);
+        }
```
lynzrand: ummm... you forgot this is the same logic too.

maybe consider factoring out into a function? I'll approve the PR because it's correct otherwise.

lynzrand (approved):

crates/moonbuild-rupes-recta/src/build_plan/builders.rs
Yu-zh: hmmm good catch...

Issue #1299, opened 2025-12-09
[BUG] NEW_MOON regression on DCE for testing


gmlewis:
**Describe the bug**
With a recent compiler update, the compilation for target `wasm-gc` is suddenly failing with no changes to the code.

**To Reproduce**
1. Clone the repo: https://github.com/gmlewis/moonbit-pdk
2. Run `./update.sh`
3. See https://github.com/gmlewis/moonbit-pdk/issues/4

**Expected behavior**
No errors.

peter-jerry-ye:
This seems to be a build system issue so I will transfer this issue there. We will investigate. In the mean time, execute with `NEW_MOON=0` will switch you back to the old build system.

peter-jerry-ye:
When testing, the build system should ignore the `export-functions` configuration from `moon.pkg.json` cc @Yu-zh

Yu-zh:
Should be fixed. Feel free to reopen if the problem still exists.

PR #1306, opened 2025-12-10
fix unexpected functions being exported when testing

Yu-zh:
- Related issues: #1299 
- PR kind: Bugfix

## Summary

When running `moon test`, it didn't ignore the exported functions provided in `moon.pkg.json`. As a result, some of public toplevels will not be eliminated and causes invalid target file.

## Metadata

### DIFF
--- a/crates/moon/tests/test_cases/no_export_when_test/mod.rs
+++ b/crates/moon/tests/test_cases/no_export_when_test/mod.rs
@@ -8,7 +8,10 @@ fn no_export_when_test() {
     compare_graphs(&build_graph, expect_file!["./build_graph.jsonl"]);
 
     let s = get_stdout(&dir, ["test"]);
-    check(s, expect![[r#"
+    check(
+        s,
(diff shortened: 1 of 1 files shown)


crates/moonbuild-rupes-recta/src/build_lower/compiler/link_core.rs
```diff
@@ -159,6 +159,8 @@ impl<'a> MooncLinkCore<'a> {
         // WebAssembly exports
         if let Some(exports) = self.exports
             && !exports.is_empty()
+            && !self.test_mode
+        // when it's test mode, we only export test driver utils
```
lynzrand, line 163: Might worth reviewing how "dumb" the commandline builders should be some time in the future. It's becoming increasingly "smart" as we're fixing bugs. 

I'm not against this, but a rule-of-thumb of "which goes to the wrappers, which goes to the args" would benefit future edits?

Issue #1304, opened 2025-12-10
Collect diagnostics from moonc and postprocess


Yu-zh:
## Motivation

1. When running `moon test`, the warnings of from a source file will be emitted twice: one for building the source package, and the other for building the whitebox tests. We hope to dedup the diagnostics.
2. We want to add more sophisticated options on the diagnostics like `--limit=10`, which will make at most 10 warnings/errors emitted, and `--filter 'path=xxx.mbt'`.

peter-jerry-ye:
Already captured. postprocess will be added on demand

PR #1309, opened 2025-12-11
collect and dedup diagnostics

Yu-zh:
- Related issues: #1304 
- PR kind: Refactor

## Summary

Previously diagnostics are printed as the command executes. This PR changes to collect them into data structures and process them before output.

Another behavioral change: legacy code has quite sophisticated logic for rendering diagnostics from the test driver. But since errors/warnings from the test driver are rare so I just skipped them and warned about errors/warnings not rendered and hint user to use `--no-render` to see the raw output.

## Metadata

### DIFF
--- a/crates/moon/tests/test_cases/diagnostics_format/test.rs
+++ b/crates/moon/tests/test_cases/diagnostics_format/test.rs
@@ -50,34 +50,6 @@ fn test_moon_test_rendered_output() {
                 │           ╰─────── Warning (unused_value): Unused variable '🤣😭🤣😭🤣'
             ────╯
             Warning: [0002]
-               ╭─[ $ROOT/lib/hello.mbt:4:7 ]
-               │
-             4 │   let a = 1;
-               │       ┬  
-               │       ╰── Warning (unused_value): Unused variable 'a'
-            ───╯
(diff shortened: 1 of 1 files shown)


crates/moon/src/rr_build/mod.rs
```diff
@@ -840,3 +842,70 @@ pub fn execute_build_partial(
...
+/// The callback function that catches diagnostics output. The output is
+/// expected to have json format. The captured diagnostics will be processed
+/// later.
+fn capture_diagnostics_callback(catcher: Arc<Mutex<ResultCatcher>>) -> impl Fn(&str) {
+    move |output: &str| {
+        output
+            .split('\n')
+            .filter(|it| !it.is_empty())
+            .for_each(|content| {
+                catcher.lock().unwrap().append_content(content, None);
```
lynzrand, line 855: Alternatives:
- Consider only accumulating diagnostic instances, and directly output the others
- Consider printing diagnostics on first sighting, and then add it to the map, so it doesn't get printed twice. (I see you're using a BTreeSet to collect, so this interferes with ordering)

lynzrand (approved):

Issue #1314, opened 2025-12-15
moon fmt crash when passed a directory


Yu-zh:
Moon formatter doesn't support formatting code within a directory like: `moon fmt cmd/main`. It should not crash at least.

## Actual behavior

```
> moon fmt cmd/main                                                                                                                                                                                master untracked
Fatal error: exception Failure("unknown file type, supported file types: mbt, md")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Moonfmt.(fun) in file "moonfmt.ml", line 131912, characters 50-70
Called from Stdlib__Arg.parse_and_expand_argv_dynamic_aux in file "arg.ml", line 279, characters 11-20
Called from Stdlib__Arg.parse in file "arg.ml", line 300, characters 4-31
Called from Moonfmt in file "moonfmt.ml", line 131911, characters 2-114
Fatal error: exception Failure("unknown file type, supported file types: mbt, md")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Moonfmt.(fun) in file "moonfmt.ml", line 131912, characters 50-70
Called from Stdlib__Arg.parse_and_expand_argv_dynamic_aux in file "arg.ml", line 279, characters 11-20
Called from Stdlib__Arg.parse in file "arg.ml", line 300, characters 4-31
Called from Moonfmt in file "moonfmt.ml", line 131911, characters 2-114
Fatal error: exception Failure("unknown file type, supported file types: mbt, md")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Moonfmt.(fun) in file "moonfmt.ml", line 131912, characters 50-70
Called from Stdlib__Arg.parse_and_expand_argv_dynamic_aux in file "arg.ml", line 279, characters 11-20
Called from Stdlib__Arg.parse in file "arg.ml", line 300, characters 4-31
Called from Moonfmt in file "moonfmt.ml", line 131911, characters 2-114
Fatal error: exception Failure("unknown file type, supported file types: mbt, md")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Moonfmt.(fun) in file "moonfmt.ml", line 131912, characters 50-70
Called from Stdlib__Arg.parse_and_expand_argv_dynamic_aux in file "arg.ml", line 279, characters 11-20
Called from Stdlib__Arg.parse in file "arg.ml", line 300, characters 4-31
[...]

PR #1318, opened 2025-12-16
allow `moon fmt` to specify a path

Yu-zh:
- Related issues: #1314 
- PR kind: Bugfix

## Summary

This PR allows `moon format <PATH>` to format the mbt files under `<PATH>`. Running command like this would make `<PATH>` treated as extra arguments that are passed to `moonfmt`, which leads to `moonfmt` crashing.

## Metadata

### DIFF
--- a/crates/moonbuild-rupes-recta/src/fmt.rs
+++ b/crates/moonbuild-rupes-recta/src/fmt.rs
@@ -132,10 +132,10 @@ pub fn build_graph_for_fmt(
 
     for &id in all_packages.values() {
         // Skip packages that don't match the filter
-        if let Some(filter_id) = package_filter {
-            if id != filter_id {
-                continue;
-            }
+        if let Some(filter_id) = package_filter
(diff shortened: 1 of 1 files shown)


crates/moon/src/cli/fmt.rs
```diff
@@ -52,6 +55,12 @@ pub(crate) struct FmtSubcommand {
     #[clap(long, conflicts_with = "check")]
     pub warn: bool,
 
+    /// Path to a package directory to format
+    #[clap(name = "PATH")]
+    pub path: Option<PathBuf>,
+
+    /// Extra arguments passed to the formatter (after --)
+    #[clap(last = true)]
     pub args: Vec<String>,
```
peter-jerry-ye, line 64: Do we really need this for `moon fmt`?

Yu-zh: idk. It was tested so I kept it.

PR #1341, opened 2025-12-29
feat: fix signal propagation from moon to child processes (closes #1337)

### LINKED ISSUE #1337: closing `moon` does not properly close `moonrun`
Create a new project

Write a simple test

```moonbit
///|
test "always" {
  let mut i = 0
  while true {
    i += 1
  }
  println(i)
}
```

Write a simple script

```js
const { spawn } = require("node:child_process");

const controller = new AbortController();
const timeoutId = setTimeout(() => controller.abort(), 5000);

const child = spawn("moon", ["test"], {
  stdio: "inherit",
  signal: controller.signal,
});

child.on("error", (err) => {
  clearTimeout(timeoutId);
  console.error(err);
  process.exit(1);
});

child.on("exit", (code, signal) => {
  clearTimeout(timeoutId);
  process.exit(code ?? 1);
});
```

Execute it. The `moonrun` will continue to exist.

peter-jerry-ye:
## Summary
- Add comprehensive signal propagation from parent moon process to child moonrun processes
- Implement signal handlers for SIGINT, SIGTERM, SIGABRT, and SIGHUP
- Create child process registry for automatic signal forwarding
- Fix issue where moonrun processes would become orphaned when moon is terminated

## Fixes
Fixes #1337 - When abort/termination signals are sent to moon process, child moonrun processes now terminate properly instead of becoming orphaned.

## Implementation
- Added `signal_handler.rs` module with thread-safe child process registry
- Setup cross-platform signal handlers using existing dependencies (`ctrlc` + `signal-hook` + `libc`)
- Modified process spawning in `run/child.rs` to use `SignalAwareChild` wrapper
- Graceful termination sequence: SIGTERM → 100ms wait → SIGKILL fallback

## Testing
- Verified signal propagation works for SIGINT, SIGTERM, and SIGABRT
- Confirmed no orphaned moonrun processes remain after signal termination
- Tested with dead loop test case from issue #1337

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,6 +1559,7 @@ dependencies = [
  "home",
  "ignore",
  "indexmap",
+ "libc",
  "log",
  "moonbuild",
  "moonbuild-debug",
@@ -1575,6 +1576,7 @@ dependencies = [
  "serde_json_lenient",
  "sha2",
  "shlex",
+ "signal-hook",
  "similar",
  "smallvec",
  "snapbox 0.6.23",
@@ -1629,6 +1631,7 @@ dependencies = [
  "serde_json",
  "serde_json_lenient",
  "shlex",
+ "signal-hook",
  "similar",
  "sysinfo",
  "tempfile",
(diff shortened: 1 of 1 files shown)


peter-jerry-ye:
Will use #1342 as base for development

Issue #1345, opened 2025-12-29
Single-file mode fails when module dependencies have package alias conflicts


Yu-zh:
## Description

Single-file mode cannot properly handle module dependencies when the dependent module contains packages with conflicting implicit aliases. This prevents users from declaring any module dependency that might have alias conflicts, effectively breaking stdlib usage in single-file mode.

### Normal package mode:

1. Declare module-level dependency in moon.mod.json
2. In each package's moon.pkg.json, explicitly import specific packages with explicit aliases
3. Build system warns about alias conflicts at the package level

### Single-file mode:

1. Declare module-level dependency in front matter
2. All packages from that module are automatically imported (implicit)
3. All aliases are implicit (derived from package names)

Problem: Alias conflicts cannot be detected or resolved because there's no package config file to report warnings to

PR #1346, opened 2025-12-29
import stdlib

Yu-zh:
- Related issues: None
- PR kind: feature

## Summary

Previously, stdlib packages were skipped during package discovery. This change includes stdlib packages in the discovery result with an `is_stdlib` flag, allowing stdlib packages to be explicitly imported by users rather than implicitly available. And automatically injects moonbitlang/core/prelude as a dependency for all non-stdlib packages. This ensures prelude definitions are available without explicit import.

This is one issue with single-file mode: https://github.com/moonbitlang/moon/issues/1345
  
## Metadata


Yu-zh:
Status: will handle alias conflicting issues in a separate PR.

Issue #806, opened 2025-05-12
moon build can't specify `cc` and `stub-cc`


illusory0x0:
# Bug Report

### OS

Operating System: Linux

### MoonBit CLI Tools Version

```
moon 0.1.20250507 (9e0b1f2 2025-05-07) ~/.moon/bin/moon
moonc v0.1.20250508+ae9fa770e ~/.moon/bin/moonc
moonrun 0.1.20250507 (9e0b1f2 2025-05-07) ~/.moon/bin/moonrun

```

## Steps to Reproduce

run `moon build  --target native  --dry-run` in [example](https://github.com/illusory0x0/bug_report.mbt/tree/master/moon-build-cc)

## Expected Behavior

I would like moon use clang to compile.

## Actual Behavior

```
moonc build-package ./src/hello.mbt -o ./target/native/release/build/hello.core -pkg username/hello -std-path /home/illu/.moon/lib/core/target/native/release/bundle -pkg-sources username/hello:./src -target native
moonc link-core /home/illu/.moon/lib/core/target/native/release/bundle/abort/abort.core /home/illu/.moon/lib/core/target/native/release/bundle/core.core ./target/native/release/build/hello.core -main username/hello -o ./target/native/release/build/hello.c -pkg-config-path ./src/moon.pkg.json -pkg-sources username/hello:./src -pkg-sources moonbitlang/core:/home/illu/.moon/lib/core -target native
gcc -o ./target/native/release/build/runtime.o -I/home/illu/.moon/include -L/home/illu/.moon/lib -g -c -fwrapv -fno-strict-aliasing -O2 /home/illu/.moon/lib/runtime.c
gcc -o ./target/native/release/build/hello.exe -I/home/illu/.moon/include -L/home/illu/.moon/lib -fwrapv -fno-strict-aliasing -O2 ./target/native/release/build/hello.c ./target/native/release/build/runtime.o -lm

```

## Checklist
- [x] (Optional) My case is minimal enough to be reproducible.

Young-Flash:
for me, it shows

```
moonc build-package ./src/hello.mbt -o ./target/native/release/build/hello.core -pkg username/hello -std-path /Users/flash/.moon/lib/core/target/native/release/bundle -pkg-sources username/hello:./src -target native
[...]

PR #1348, opened 2025-12-29
feat: warn when both MOON_CC and user CC are provided

Yu-zh:
## Summary

Add a warning when both the MOON_CC environment variable and a user-specified CC (from moon.pkg.json) are provided, since MOON_CC takes precedence and the user-specified CC will be ignored.


Yu-zh:
Related: https://github.com/moonbitlang/moon/issues/806#issuecomment-2876363401

PR #1415, opened 2026-01-28
allow using symlinked files as sources

Yu-zh:
- Related issues: #1413
- PR kind:  Bugfix

## Summary

Updated discovery to treat symlinked files as eligible source entries (matching the legacy scanner’s behavior), so .mbt/.mbt.md symlinks in a package dir are now picked up.

Notes

- This doesn’t change directory traversal: symlinked dirs are still skipped by the WalkDir scan.
- Only the symlink’s own name/extension is used for inclusion, same as before.

### DIFF
--- a/crates/moon/tests/test_cases/symlink_file_discovery/mod.rs
+++ b/crates/moon/tests/test_cases/symlink_file_discovery/mod.rs
@@ -13,6 +13,7 @@ fn test_symlink_file_discovery() {
     check(
         get_stdout(&dir, ["run", "main"]),
         expect![[r#"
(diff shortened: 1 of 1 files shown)


Yu-zh:
@peter-jerry-ye What should happen if running `moon publish` with symlinked source files?

peter-jerry-ye:
The publish will (in the future) pack the zip and unzip in another place to verify. If the symlink doesn't break in such case, then it shall be fine.

peter-jerry-ye:
@copilot Add test where the symlink is recursive

PR #1416, opened 2026-01-28
support backtrace

Yu-zh:


Yu-zh:
It seems only worked on Mac machine @Kaida-Amethyst

PR #1417, opened 2026-01-28
feat: support new moon install and deprecate old usage

Lampese:
The old moon install behavior has been abandoned and warnings have been added. Now, using `moon install xxx/yyy` will search for all packages in the package with `is main` true and install them to `.moon/mooncakes_bin`(must be native packages, if `moon check --target native` fails, stop the installation). Afterwards, they can be called through `exename`.

We also support a option `--path` to install local modules.

## Metadata


crates/moon/src/cli/deps.rs
```diff
@@ -16,26 +16,297 @@
...
+        bail!("--dry-run is not supported for `moon install <package>`");
+    }
+
+    let (pkg_name, version) = resolve_package_version(&cli, package_path)?;
+    let temp_dir = TempDir::new().context("failed to create temp directory for install")?;
+    let pkg_dir = temp_dir.path().join("pkg");
+
+    let registry = OnlineRegistry::mooncakes_io();
+    registry.install_to(&pkg_name, &version, &pkg_dir, cli.quiet)?;
+
```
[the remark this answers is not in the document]
Lampese: This path currently depends on OnlineRegistry (index + package zip from mooncakes.io) and then triggers a native build. A stable test would require an offline registry/index fixture plus a cached zip with matching checksum, and also a working MoonBit toolchain/native backend in CI. That’s a lot of moving parts and tends to make tests flaky across environments. We already have CI‑only smoke tests for online flows, so I’d prefer to defer coverage until we can inject a mock registry or add a deterministic offline fixture.

crates/moon/src/cli/deps.rs
```diff
@@ -16,26 +16,297 @@
...
+            .clone()
+            .unwrap();
+        if !cli.quiet {
+            println!("Latest version of {pkg_name} is {latest_version}");
+        }
+        latest_version
+    };
+
+    Ok((pkg_name, version))
+}
```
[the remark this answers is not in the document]
Lampese: I agree there’s duplication, but extracting a shared helper would touch add/fetch/install and likely change messages/behavior in those commands. That’s a larger refactor than this PR’s scope. I’d prefer to land the behavior change first and follow up with a dedicated refactor PR to unify parsing + index update logic.

docs/manual-zh/src/commands.md
```diff
@@ -423,9 +423,15 @@ Remove a dependency
 
 ## `moon install`
 
-Install dependencies
+Note: `moon install` without `PACKAGE_PATH` is deprecated; use `moon build` instead.
```
peter-jerry-ye:
```suggestion
Note: `moon install` without `PACKAGE_PATH` is deprecated; use `moon check` or `moon build` directly.
```

crates/moon/src/cli/deps.rs
```diff
@@ -16,26 +16,297 @@
 //
 // For inquiries, you can contact us via e-mail at jichuruanjian@idea.edu.cn.
 
-use anyhow::bail;
+use anyhow::{Context as _, bail};
```
peter-jerry-ye: Do we need context or not?

crates/moon/src/cli/deps.rs
```diff
@@ -16,26 +16,297 @@
...
+        let expected = format!(".{ext}");
+        let keep_ext = !ext.eq_ignore_ascii_case("exe") || cfg!(windows);
+        if keep_ext && !file_name.ends_with(&expected) {
+            file_name.push_str(&expected);
+        }
+    }
+    if !cfg!(windows) && file_name.to_ascii_lowercase().ends_with(".exe") {
+        let new_len = file_name.len() - 4;
+        file_name.truncate(new_len);
+    }
```
peter-jerry-ye: Simplify `.exe` handling

docs/manual-zh/src/commands.md
Lampese: updated

crates/moon/src/cli/deps.rs
Lampese: Yes, `as _` explicitly discards the imported name while still bringing the trait into scope.
such as:
```rust
use std::io::Read as _;
let mut s = String::new();
std::fs::File::open("a.txt")?.read_to_string(&mut s)?;
```

crates/moon/tests/install_package.rs
```diff
@@ -0,0 +1,284 @@
...
+        std::os::unix::fs::symlink(&real_lib, temp.path().join("lib")).ok()?;
+        std::os::unix::fs::symlink(&real_include, temp.path().join("include")).ok()?;
+    }
+    #[cfg(windows)]
+    {
+        // Avoid Windows symlink permission issues for now.
+        return None;
+    }
+    Some(temp)
+}
```
peter-jerry-ye: Why is this so complicated? Should just spawn process with `MOON_HOME` set to a temp dir for testing.

crates/moon/src/cli.rs
```diff
@@ -457,3 +458,19 @@ fn gen_docs_for_moon_help_page() {
...
+        let insert_at = pos + NEEDLE.len();
+        let mut out = String::with_capacity(markdown.len() + NOTE.len());
+        out.push_str(&markdown[..insert_at]);
+        out.push_str(NOTE);
+        out.push_str(&markdown[insert_at..]);
+        out
+    } else {
+        markdown.to_string()
+    }
+}
```
peter-jerry-ye: What's the use of this block? And why `cfg(test)`?

crates/moon/src/cli/external.rs
peter-jerry-ye, line 1: Why is this file changed?

crates/moon/src/cli/deps.rs
peter-jerry-ye, line 1: Should create another file because it's no longer related to dependency management.

peter-jerry-ye (changes requested):

PR #1422, opened 2026-01-29
Update symbols cache on moon update

bzy-debug:
## Summary

- download and refresh symbols.zip during `moon update`
- add a shared zip extraction helper for symbols and registry installs

symbols.zip contains pre-computed symbols of package moonbitlang/async and moonbitlang/x, and is used for `moon ide doc -registry` for symbol search

after `moon update`, there should be a `symbols` folder in `~/.moon/registry` with content:

```
symbols/
└── moonbitlang
    ├── async
    │   └── 0.16.3.symbols
    └── x
        └── 0.4.38.symbols

4 directories, 2 files
```


bzy-debug:
@codex review

Issue #1359, opened 2026-01-05
Regression: complicated js test


peter-jerry-ye:
The old moon uses `node target/js/xxx/xxx.cjs "{}"`  to execute the tests. But the new moon moves the driver into some arbitrary directory, making it literally impossible to rerun the node command to adjust parameters to to debug js code.

peter-jerry-ye:
To fix this problem properly, there should exist a stable and fixed javascript file that serves as the entrance of the testing, which is responsible for parsing and calling the execution of files containing tests.

PR #1425, opened 2026-01-29
refactor: not put js test driver in tmp file

Yu-zh:
- Related issues: #1359
- PR kind: refactor

## Summary

The new backend generates the `driver.cjs` into a temporary dir so that the `driver.cjs` is properly cleaned up when the test finishes. But this making debug with the driver hard. So this PR changes it to store the driver in the build target dir.


peter-jerry-ye:
We will refactor and merge the js test driver with the native test driver, so this pr I think is no longer necessary. And with that refactor, we will also allow having `.cjs` or `mjs` following the package configuration.

PR #1430, opened 2026-02-01
feat: add new moon install and deprecate old usage

Lampese:
This PR implements `moon install <package>` command for installing executable packages globally, similar to Go's `go install`. Users can install packages from the Mooncakes registry using a path like `user/module/main` or specify a version with `@version` suffix. The command also supports wildcard patterns (`user/module/...`) to install all main packages in a module, and local installation via `--path` flag for development workflows. Installed binaries are placed in `~/.moon/bin/` by default (or a custom directory via `--bin`), making them available system-wide.

### DIFF
--- a/crates/moonutil/src/moon_dir.rs
+++ b/crates/moonutil/src/moon_dir.rs
@@ -121,6 +121,7 @@ pub const RESERVED_BIN_NAMES: &[&str] = &[
     "moonrun",
     "moon_cove_report",
     "moon-ide",
+    "moon-lsp",
     "moon-pilot",
     "moon-wasm-opt",
     "moonbit-lsp",


crates/moon/src/cli/install_binary.rs
```diff
@@ -0,0 +1,416 @@
...
+    if components.len() < 2 {
+        bail!(
+            "Invalid package path `{}`: must be in format `user/module/package`",
+            input
+        );
+    }
+
+    let module_name = ModuleName {
+        username: components[0].into(),
+        unqual: components[1].into(),
```
peter-jerry-ye, line 81: What is unqual?

peter-jerry-ye (approved):

crates/moon/src/cli/install_binary.rs
Lampese: unqualified
‌It is the same here:
image

PR #1433, opened 2026-02-02
fix coverage on JavaScript backend for async test

Guest0x0:
Coverage is not working properly on JavaScript backend when `async test` is present. `@coverage.end()` should be called after all tests are run to properly collect coverage data. Previously, the JS test driver call `@coverage.end()` at the end of the entry script `js_driver.js`. However, if `async test` is present, there may be promises created by `async test` running in the background at this point, and coverage data for those tests would be missing.

This PR fixes the problem by always calling `@coverage.end()` after all `async test` completed. Note that `@coverage.end()` will reset the global counter and can be safely invoked multiple times, so no need to change other parts of code.

Duplicated call to `@coverage.end()` may be avoidable , but that's out of the scope of this PR.

I've tested locally and this PR can indeed produce correct coverage data on `moonbitlang/async`.


peter-jerry-ye:
Shall we remove the other coverage end?

PR #1435, opened 2026-02-03
feat: add git source and add path recognition for moon install

Lampese:
This PR adds `--git` support to `moon install`, allowing users to install packages directly from remote git repositories. Usage includes `moon install --git <url>` for installing root packages, `moon install --git <url> cmd/main` for specific packages, and `moon install --git <url> cmd/...` for wildcard patterns. Optional `--branch`, `--tag`, or `--rev` flags allow specifying the checkout target.

### DIFF
--- a/crates/moon/src/cli/deps.rs
+++ b/crates/moon/src/cli/deps.rs
@@ -64,15 +64,20 @@ pub fn install_cli(cli: UniversalFlags, cmd: InstallSubcommand) -> anyhow::Resul
 
     let package_path = cmd.package_path.unwrap();
 
-    // Auto-detect local paths
+    // Local path install
+    // These checks can't be done in clap because we need to inspect the value of package_path
+    // to determine whether it's a local path, git URL, or registry path.
     if is_local_path(&package_path) {
         if has_git_ref {
             anyhow::bail!("--rev, --branch, and --tag can only be used with git URLs");
         }
+        if cmd.package_path_in_repo.is_some() {
+            anyhow::bail!("Package path in repo can only be used with git URLs");
+        }
         return install_from_local(&cli, package_path.as_ref(), &install_dir);
     }
 
-    // Auto-detect git URLs
+    // Git URL install
     if is_git_url(&package_path) {
         let git_ref = if let Some(rev) = cmd.rev.as_deref() {
             GitRef::Rev(rev)
@@ -83,12 +88,21 @@ pub fn install_cli(cli: UniversalFlags, cmd: InstallSubcommand) -> anyhow::Resul
         } else {
             GitRef::Default
         };
-        return install_from_git(&cli, &package_path, git_ref, None, &install_dir);
+        return install_from_git(
+            &cli,
+            &package_path,
+            git_ref,
+            cmd.package_path_in_repo.as_deref(),
+            &install_dir,
+        );
     }
 
     // Registry install
     if has_git_ref {
         anyhow::bail!("--rev, --branch, and --tag can only be used with git URLs");
+    }
+    if cmd.package_path_in_repo.is_some() {
+        anyhow::bail!("Package path in repo can only be used with git URLs");
     }
     let spec = parse_package_spec(&package_path)?;
     install_binary(&cli, &spec, &install_dir)
--- a/crates/moon/src/cli/install_binary.rs
+++ b/crates/moon/src/cli/install_binary.rs
@@ -339,7 +339,12 @@ pub(super) fn install_from_git(
                 .to_string(),
         }
     } else {
-        PackageFilter::ByPath(target_path)
+        // Use ByPackagePath for git install (string comparison, not filesystem path)
+        PackageFilter::ByPackagePath(
+            package_path
+                .map(|p| p.trim_matches('/').to_string())
+                .unwrap_or_default(),
+        )
     };
 
     build_and_install_packages(cli, &module_name, &module_root, install_dir, filter)
--- a/crates/mooncake/src/pkg/install.rs
+++ b/crates/mooncake/src/pkg/install.rs
@@ -46,12 +46,21 @@ pub struct InstallSubcommand {
     /// If not provided, falls back to legacy behavior (install project dependencies)
     pub package_path: Option<String>,
 
+    /// Package path within a git repository (e.g., src/main or cmd/...)
+    /// Only used with git URLs. Supports /... suffix for wildcard matching.
+    pub package_path_in_repo: Option<String>,
+
     /// Specify installation directory (default: ~/.moon/bin/)
     #[clap(long)]
     pub bin: Option<PathBuf>,
 
     /// Install from local path instead of registry
-    #[clap(long, conflicts_with = "package_path", conflicts_with = "git_ref")]
+    #[clap(
+        long,
+        conflicts_with = "package_path",
+        conflicts_with = "git_ref",
+        conflicts_with = "package_path_in_repo"
+    )]
     pub path: Option<PathBuf>,
 
     /// Git revision to checkout (commit hash, requires git URL)
--- a/crates/moon/tests/test_cases/mod.rs
+++ b/crates/moon/tests/test_cases/mod.rs
@@ -2972,7 +2972,6 @@ fn test_moon_install_global_deprecated_warning() {
 }
 
 #[test]
-#[ignore = "requires native backend support"]
 fn test_moon_install_global_local_path() {
     // Test installing from local path using --path
     let dir = TestDir::new("moon_install_global.in");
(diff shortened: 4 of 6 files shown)


crates/moon/src/cli/install_binary.rs
```diff
@@ -201,6 +208,134 @@ pub fn install_from_local(
...
+
+    if !quiet {
+        eprintln!("{}: Cloning `{}`...", "Info".cyan(), git_url);
+    }
+
+    let tmp_dir = tempfile::TempDir::new().context("Failed to create temporary directory")?;
+    let clone_dir = tmp_dir.path();
+
+    // Clone the repository
+    let mut clone_cmd = std::process::Command::new("git");
```
peter-jerry-ye: Use `BINARIES.git`

crates/moon/src/cli/install_binary.rs
```diff
@@ -201,6 +208,134 @@ pub fn install_from_local(
...
+    match git_ref {
+        GitRef::Branch(branch) => {
+            clone_cmd.arg("--branch").arg(branch);
+        }
+        GitRef::Tag(tag) => {
+            clone_cmd.arg("--branch").arg(tag);
+        }
+        GitRef::Rev(_) => {
+            // For rev, we need full clone then checkout
+            clone_cmd.args(["--depth", "1"]).arg("--no-checkout");
```
peter-jerry-ye: Looks like contradicting.

crates/moon/src/cli/install_binary.rs
```diff
@@ -201,6 +208,134 @@ pub fn install_from_local(
...
+
+    clone_cmd.arg(git_url).arg(clone_dir);
+
+    let status = clone_cmd.status().context("Failed to execute git clone")?;
+
+    if !status.success() {
+        bail!("Failed to clone repository `{}`", git_url);
+    }
+
+    // If rev specified, fetch and checkout
```
peter-jerry-ye: `--revision`?

crates/mooncake/src/pkg/install.rs
```diff
@@ -48,8 +50,24 @@ pub struct InstallSubcommand {
...
+    pub rev: Option<String>,
+
+    /// Git branch to checkout
+    #[clap(long, requires = "git", conflicts_with_all = ["rev", "tag"])]
+    pub branch: Option<String>,
+
+    /// Git tag to checkout
+    #[clap(long, requires = "git", conflicts_with_all = ["rev", "branch"])]
+    pub tag: Option<String>,
 }
```
peter-jerry-ye: Use ArgGroup with `multiple = false`

crates/mooncake/src/pkg/install.rs
```diff
@@ -48,8 +50,24 @@ pub struct InstallSubcommand {
     pub bin: Option<PathBuf>,
 
     /// Install from local path instead of registry
-    #[clap(long, conflicts_with = "package_path")]
+    #[clap(long, conflicts_with_all = ["package_path", "git"])]
     pub path: Option<PathBuf>,
+
+    /// Install from a git repository URL
+    #[clap(long, conflicts_with = "path")]
+    pub git: Option<String>,
```
peter-jerry-ye: Use ArgGroup with multiple = false

crates/moon/src/cli/install_binary.rs
```diff
@@ -47,6 +47,13 @@ pub struct PackageSpec {
     pub is_wildcard: bool,
 }
 
+const GIT_URL_PREFIXES: &[&str] = &["https://", "http://", "git://", "ssh://", "file://", "git@"];
+
+/// Check if a string looks like a git URL.
+pub fn is_git_url(s: &str) -> bool {
+    GIT_URL_PREFIXES.iter().any(|p| s.starts_with(p))
+}
```
peter-jerry-ye: `file://` treated as git here is dangerous.

crates/moon/src/cli/install_binary.rs
```diff
@@ -201,6 +208,134 @@ pub fn install_from_local(
...
+
+    // Determine if wildcard or specific package
+    let is_module_root = target_path == module_root;
+    let is_wildcard = package_path
+        .map(|p| p.ends_with("/...") || p == "...")
+        .unwrap_or(is_module_root);
+
+    let spec = PackageSpec {
+        module_name,
+        package_path: Some(String::new()),
```
peter-jerry-ye: Are you sure about this?

crates/moon/src/cli/install_binary.rs
Lampese: cargo is --rev, we can use --revision

crates/moon/src/cli/install_binary.rs
Lampese: Yes, this field isn't used here. Filtering happens via filter_by_path instead.

Maybe I need refactor it?

crates/moon/src/cli/install_binary.rs
Lampese: Ok, I'll remove it

crates/moon/src/cli/install_binary.rs
peter-jerry-ye: I meant `git clone --revision`

crates/moon/src/cli/install_binary.rs
peter-jerry-ye: Yes, make it clearer. Maybe use an enum.

crates/moon/src/cli/install_binary.rs
Lampese: https://lore.kernel.org/git/xmqqfrjfilc8.fsf@gitster.g/

This parameter was only introduced in the first half of 2025, and I think many old server machines may not have pre installed git that supports it

peter-jerry-ye: Okey, we might postpone this or we might vendor the libgit2 later.

Lampese: By the way, be careful of the volume pressure caused by libgit2. I remember I once introduced it for a small feature of moon, but it resulted in an increase in binary volume compared to the original binary

Lampese:
@peter-jerry-ye 

New behavior update: The flag `--git` has been removed and replaced with automatic recognition of git links, which can also be automatically recognized local paths start with `./`(--path still retained).

Other: Added testing for parsing section

crates/moon/src/cli/deps.rs
```diff
@@ -51,18 +53,47 @@ pub fn install_cli(cli: UniversalFlags, cmd: InstallSubcommand) -> anyhow::Resul
...
-        // Install from local path
-        install_from_local(&cli, &local_path, &install_dir)
-    } else {
-        // Install from registry
-        let package_path = cmd.package_path.unwrap();
-        let spec = parse_package_spec(&package_path)?;
-        install_binary(&cli, &spec, &install_dir)
+        if has_git_ref {
+            anyhow::bail!("--rev, --branch, and --tag can only be used with git URLs");
+        }
```
peter-jerry-ye: Configure clap for this?

crates/moon/src/cli/deps.rs
```diff
@@ -51,18 +53,47 @@ pub fn install_cli(cli: UniversalFlags, cmd: InstallSubcommand) -> anyhow::Resul
...
+
+    let package_path = cmd.package_path.unwrap();
+
+    // Auto-detect local paths starting with ./
+    if package_path.starts_with("./") {
+        if has_git_ref {
+            anyhow::bail!("--rev, --branch, and --tag can only be used with git URLs");
+        }
+        return install_from_local(&cli, package_path.as_ref(), &install_dir);
+    }
```
peter-jerry-ye: Also for absolute path

crates/moon/src/cli/install_binary.rs
```diff
@@ -47,6 +47,23 @@ pub struct PackageSpec {
...
+    /// Match all main packages, optionally with a prefix (for wildcard patterns)
+    Wildcard { prefix: String },
+    /// Match by package path string (for registry install)
+    ByPackagePath(String),
+}
+
+const GIT_URL_PREFIXES: &[&str] = &["https://", "http://", "git://", "ssh://", "git@"];
+
+/// Check if a string looks like a git URL.
+pub fn is_git_url(s: &str) -> bool {
```
peter-jerry-ye: Unnecessary `pub`.

crates/moon/src/cli/install_binary.rs
```diff
@@ -184,33 +209,140 @@ pub fn install_from_local(
...
+    /// Checkout a branch
+    Branch(&'a str),
+    /// Checkout a tag
+    Tag(&'a str),
+    /// Use default branch
+    Default,
+}
+
+/// Install from a git repository.
+pub fn install_from_git(
```
peter-jerry-ye: Avoid unnecessary `pub`

crates/moon/src/cli/deps.rs
Lampese: So actually, we still need to support `/` `../`. If it's on Windows, we may also need to consider determining the drive letter. I have already supported common situations.

crates/moon/src/cli/install_binary.rs
Lampese: Also used in deps.rs, changed to pub(super).

crates/moon/src/cli/install_binary.rs
Lampese: Also used in deps.rs, changed to pub(super).

Lampese:
> We also need one test

The testing for directory parsing has been added to `install_binary.rs` last time. For the testing of git, I am not sure which testing method is reliable because currently we only support installing packages from online sources.

peter-jerry-ye (approved):

PR #1450, opened 2026-02-05
moon ide shell completion

bzy-debug:
This MR adds shell completion for `moon ide`

Since `moon ide` is not really a subcommand, so we only register `moon ide`'s args on the fly before `moon shell-completion`.


bzy-debug:
@codex review

PR #1459, opened 2026-02-09
fix: not use `target` in bundle core path

Yu-zh:


crates/moonbuild-rupes-recta/src/build_lower/artifact.rs
```diff
@@ -569,7 +569,7 @@ fn object_file_ext(os: OperatingSystem) -> &'static str {
 /// removed in the future.
 pub fn core_bundle_path(core_root: &Path, backend: TargetBackend) -> PathBuf {
     let mut path = PathBuf::from(core_root);
-    path.push("target");
+    path.push("_build");
```
peter-jerry-ye: Should use `moonutil::BUILD_DIR`

PR #1468, opened 2026-02-10
refactor: cleanup unnecessary clone

Yu-zh:
- Replace HashMap double-lookups with .entry() API
- Add with_capacity() to HashMaps where size is known
- Replace indexed loops with .enumerate()
- Eliminate chained String clones in scan.rs warn_list/alert_list
- Avoid HashMap clone in MVS resolver by iterating with .chain()
- Replace iterator clone for emptiness check with direct .is_empty()
- Change resolve_cc to accept references, removing clones at call sites
- Use .to_string_lossy().into_owned() instead of .to_string_lossy().to_string()
- Replace split(" ") with split_whitespace()
- Hoist version.to_string() out of loop in registry lookup
- Use format!() instead of String + concatenation
- Use .clone() instead of .to_string() on &String in sync.rs
- Hoist pkg_name.to_string() to local variable in add.rs

### DIFF
--- a/crates/mooncake/src/pkg/sync.rs
+++ b/crates/mooncake/src/pkg/sync.rs
@@ -63,7 +63,7 @@ pub fn auto_sync_for_single_mbt_md(
         front_matter_config.and_then(|config| config.moonbit.unwrap_or_default().deps)
     {
         for (k, v) in deps_map.iter() {
-            deps.insert(k.to_string(), v.clone().into());
+            deps.insert(k.clone(), v.clone().into());
         }
     }
 
@@ -99,7 +99,7 @@ pub fn auto_sync_for_single_file_rr(
     let mut synth_deps = IndexMap::new();
     if let Some(deps_map) = front_matter_deps {
         for (k, v) in deps_map.iter() {
-            synth_deps.insert(k.to_string(), v.clone().into());
+            synth_deps.insert(k.clone(), v.clone().into());
         }
     }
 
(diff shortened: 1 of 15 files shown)


crates/mooncake/src/registry.rs
```diff
@@ -98,7 +98,7 @@ impl RegistryList {
     }
 
     pub fn with_registry(registry: Box<dyn Registry>) -> Self {
-        let mut registries = HashMap::new();
+        let mut registries = HashMap::with_capacity(1);
```
peter-jerry-ye, line 101: Is this really better?

crates/mooncake/src/registry/online.rs
```diff
@@ -141,9 +141,10 @@ impl OnlineRegistry {
         let reader = std::io::BufReader::new(file);
 
         let lines = reader.lines().collect::<std::io::Result<Vec<String>>>()?;
+        let version_str = version.to_string();
         for line in lines.iter().rev() {
             let j: MoonModJSON = serde_json_lenient::from_str(line)?;
-            if j.version.as_ref() == Some(&version.to_string()) {
+            if j.version.as_ref() == Some(&version_str) {
```
peter-jerry-ye, line 147: Is this better?

crates/mooncake/src/pkg/sync.rs
peter-jerry-ye, line 1: Is this necessary?

crates/moonbuild/src/gen/gen_runtest.rs
peter-jerry-ye, line 1: Is this necessary better?

crates/moonbuild-rupes-recta/src/build_plan/builders.rs
peter-jerry-ye, line 1: Is this better?

PR #1479, opened 2026-02-12
feat: moon run source.mbtx support

Yoorkin:
- Related issues: None 
- PR kind: feature

## Summary

This is a feature we want to experiment with in the nightly channel.

Note: I made some changes to ensure that `_build` and `.mooncakes` are generated in a temporary workspace, to avoid modifying the user's source directory. Currently, this should not affect `.mbt.md`.

```
moon run server.mbtx --target native
```

server.mbtx:

```moonbit
// comment
import {
  "moonbitlang/core/env",
  "moonbitlang/async@0.16.6",
  "moonbitlang/async@0.16.6/socket",
  "moonbitlang/async@0.16.6/fs",
  "moonbitlang/async@0.16.6/http",
  "moonbitlang/async@0.16.6/process" @process,
}

///|
async fn serve_directory(
  conn : @http.ServerConnection,
  dir : @fs.Directory,
  path~ : String,
) -> Unit {
  let files = dir.read_all()
  files.sort()
  conn
  ..send_response(200, "OK", extra_headers={ "Content-Type": "text/html" })
  ..write("<!DOCTYPE html><html><head></head><body>")
  ..write("<h1>\{path}</h1>\n")
  ..write("<div style=\"margin: 1em; font-size: 15pt\">")
  .write("<a href=\"\{path}?download_zip\">download as zip</a><br/><br/>\n")
  if path[:-1].rev_find("/") is Some(index) {
    let parent : StringView = if index == 0 { "/" } else { path[:index] }
    conn.write("<a href=\"\{parent}\">..</a><br/><br/>\n")
  }
  for file in files {
    let sep = if path is [.., '/'] { "" } else { "/" }
    conn.write("<a href=\"\{path}\{sep}\{file}\">\{file}</a><br/>\n")
  }
  conn..write("</div></body></html>").end_response()
}

///|
async fn serve_file(
  conn : @http.ServerConnection,
  file_path : String,
  path~ : String,
) -> Unit {
  let file = @fs.open(file_path, mode=ReadOnly)
  defer file.close()
  let content_type = match path {
    [.., .. ".png"] => "image/png"
    [.., .. ".jpg"] | "jpeg" => "image/jpeg"
    [.., .. ".html"] => "text/html"
    [.., .. ".css"] => "text/css"
    [.., .. ".js"] => "text/javascript"
    [.., .. ".mp4"] => "video/mp4"
    [.., .. ".mpv"] => "video/mpv"
    [.., .. ".mpeg"] => "video/mpeg"
    [.., .. ".mkv"] => "video/x-matroska"
    _ => "appliaction/octet-stream"
  }
  conn
  ..send_response(200, "OK", extra_headers={ "Content-Type": content_type })
  ..write_reader(file)
  .end_response()
}

///|
#cfg(platform="windows")
let is_windows : Bool = true

///|
#cfg(not(platform="windows"))
let is_windows : Bool = false

///|
async fn serve_zip(conn : @http.ServerConnection, path : String) -> Unit {
  let full_path = @fs.realpath(path)
  let path_sep = if is_windows { "\\" } else { "/" }
  let base_name = if full_path.rev_find(path_sep) is Some(i) {
    full_path[i + 1:].to_string()
  } else {
    path
  }
  @async.with_task_group(group => {
    let (we_read_from_zip, zip_write_to_us) = @process.read_from_process()
    defer we_read_from_zip.close()
    if is_windows {
      @process.spawn(
        group,
        "tar",
        ["-c", "-z", "-f", "-", path],
        stdout=zip_write_to_us,
      )
      |> ignore
    } else {
      @process.spawn(
        group,
        "zip",
        ["-q", "-r", "-", path],
        stdout=zip_write_to_us,
      )
      |> ignore
    }
    conn
    ..send_response(200, "OK", extra_headers={
      "Content-Type": "application/octet-stream",
      "Content-Disposition": "filename=\{base_name}.zip",
    })
    ..write_reader(we_read_from_zip)
    .end_response()
  })
}

///|
let page_for_404 : Bytes = b"<html><head></head><body>Page Not Found</body></html>"

///|
async fn serve_404(conn : @http.ServerConnection) -> Unit {
  conn
  ..send_response(404, "NotFound", extra_headers={ "Content-Type": "text/html" })
  ..write(page_for_404)
  .end_response()
}

///|
pub async fn server_main(
  server : @http.Server,
  path~ : String,
  log? : (String) -> Unit = println,
) -> Unit {
  let base_path = path
  server.run_forever((request, _body, conn) => {
    let (path, download_zip) = match request.path {
      [.. path, .. "?download_zip"] => (path.to_string(), true)
      path => (path, false)
    }
    log("serving \{path}")
    guard request.meth is Get else { return }
    let full_path = base_path + path
    guard @fs.exists(full_path) else {
      serve_404(conn)
      return
    }
    if @fs.kind(full_path) is Directory {
      if download_zip {
        serve_zip(conn, full_path)
      } else {
        let dir = @fs.opendir(full_path)
        defer dir.close()
        serve_directory(conn, dir, path~)
      }
    } else {
      serve_file(conn, full_path, path~)
    }
  })
}

///|
async fn main {
  println("hello")
  let path = match @env.args() {
    [] | [_] => "."
    [_, path, ..] => path
  }
  let server = @http.Server::new(@socket.Addr::parse("0.0.0.0:8200"))
  server_main(server, path~) catch {
    err => println("server terminate due to \{err}")
  }
}
```

## Metadata

### DIFF
--- a/crates/moon/src/cli/check.rs
+++ b/crates/moon/src/cli/check.rs
@@ -187,6 +187,7 @@ fn run_check_for_single_file_rr(
     );
     let (resolved, backend) = moonbuild_rupes_recta::resolve::resolve_single_file_project(
         &resolve_cfg,
+        raw_target_dir.as_path(),
         &single_file_path,
         false,
     )?;
--- a/crates/moon/src/cli/run.rs
+++ b/crates/moon/src/cli/run.rs
@@ -58,26 +58,30 @@ pub(crate) struct RunSubcommand {
 
 #[instrument(skip_all)]
 pub(crate) fn run_run(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Result<i32> {
+    let input = cmd.package_or_mbt_file.as_str();
+    let is_mbt = input.ends_with(".mbt");
+    let is_mbtx = input.ends_with(".mbtx");
+
     match cli.source_tgt_dir.try_into_package_dirs() {
         Ok(_) => {
-            if cmd.package_or_mbt_file.ends_with(".mbt") {
+            if is_mbtx {
+                return run_single_file_rr(cli, cmd);
+            }
+            if is_mbt {
                 let moon_pkg_json_exist = std::env::current_dir()?
-                    .join(&cmd.package_or_mbt_file)
+                    .join(input)
                     .parent()
                     .is_some_and(is_moon_pkg_exist);
                 if !moon_pkg_json_exist {
                     return run_single_file_rr(cli, cmd);
                 }
             }
-            // moon should report an error later if the source_dir doesn't
-            // contain moon.pkg.json
         }
         Err(e @ moonutil::dirs::PackageDirsError::NotInProject(_)) => {
-            if cmd.package_or_mbt_file.ends_with(".mbt") {
+            if is_mbt || is_mbtx {
                 return run_single_file_rr(cli, cmd);
-            } else {
-                return Err(e.into());
             }
+            return Err(e.into());
         }
         Err(e) => return Err(e.into()),
     }
@@ -232,6 +236,7 @@ fn run_single_file_rr(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Resul
     );
     let (resolved, backend) = moonbuild_rupes_recta::resolve::resolve_single_file_project(
         &resolve_cfg,
+        raw_target_dir.as_path(),
         &input_path,
         true,
     )?;
--- a/crates/moon/src/cli/test.rs
+++ b/crates/moon/src/cli/test.rs
@@ -322,6 +322,7 @@ fn run_test_in_single_file_rr(cli: &UniversalFlags, cmd: &TestSubcommand) -> any
     );
     let (resolved, backend) = moonbuild_rupes_recta::resolve::resolve_single_file_project(
         &resolve_cfg,
+        raw_target_dir.as_path(),
         &single_file_path,
         false,
     )?;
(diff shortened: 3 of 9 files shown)


Yu-zh:
Below is the comment from codex and I think this should be taken seriously.

---

The new .mbtx path handling miscomputes the temporary workspace when --target-dir is set, leading to writes outside the configured build directory and permission
  failures for common target paths. This is a functional regression that should be fixed before considering the patch correct.

  Review comment:

  - [P1] Keep .mbtx workspace inside configured target dir — moonbuild/crates/moon/src/cli/run.rs:233-236
    For .mbtx runs, temp_workspace is derived from target_dir.parent(), so a custom --target-dir causes generated files (<stem>.mbt and .mooncakes) to be written
    outside the chosen build directory. This breaks valid configurations like --target-dir /tmp (workspace becomes /private//, often not writable) and can silently
    pollute unrelated directories when --target-dir is relative. The workspace should be rooted under the target directory (or a dedicated subdir), not its parent.

Yoorkin:
@Yu-zh 

It seems .mbt.md mode has ignored `--target-dir` historically. If we want to respect `--target-dir` consistently across all single-file modes, we should probably redesign the default value of `--target-dir` based on input type:

  1. .mbt: default to <project_root>/_build
  2. .mbt.md: default to `<tmp>/<random_dir>/_build` 
  3. .mbtx: default to `<tmp>/<random_dir>/_build`

We could also introduce a hidden flag like `--deps-cache-dir` to control where .mooncakes is stored.

Since hiding build/cache files requires careful handling for both `.mbt.md` and `.mbtx`, I removed those related changes to keep this PR focused on making `moon run server.mbtx --target native` work. For now, .mooncakes and _build are still generated next to the source file, consistent with current .mbt.md behavior.

PR #1483, opened 2026-02-21
refactor: remove unused result

Yu-zh:


Yu-zh:
The test for abort trace failed and I skipped it for now @peter-jerry-ye

PR #1497, opened 2026-03-03
fix(install): clarify source/path semantics and improve install help

peter-jerry-ye:
## Summary
- rename install positional args to `SOURCE` and `PATH_IN_REPO` to reflect semantics
- keep single-source guessing but make semantics explicit:
  - local path markers => filesystem path mode
  - git URL => git mode
  - otherwise => registry package path mode
- clarify git second argument as filesystem path inside repo (not logical package path)
- improve `moon install --help` readability with structured clap `help`/`long_help` and value names
- add and link a dev reference page documenting expected binary installer behavior
- document caveats for ambiguous relative selectors and nested-module git repos

## Validation
- `cargo test -p moon install_binary`
- `cargo run -q -p moon -- install --help`


peter-jerry-ye:
Split to #1501 and #1502

PR #1538, opened 2026-03-19
refactor: remove unused registry config plumbing

peter-jerry-ye:
## Summary
- remove unused RegistryConfig plumbing from RR resolve and sync entry points
- keep registry configuration limited to index update paths and document the current split between update, MVS resolution, and downloads
- clean up stale resolver comments after the API simplification

## Testing
- cargo test -p moon --test add_updates_index
- cargo test -p mooncake resolver::mvs --lib

### DIFF
--- a/crates/mooncake/src/pkg/remove.rs
+++ b/crates/mooncake/src/pkg/remove.rs
@@ -22,7 +22,6 @@ use std::{path::Path, sync::Arc};
 use moonutil::{
     common::{read_module_desc_file_in_dir, write_module_json_to_file},
     module::convert_module_to_mod_json,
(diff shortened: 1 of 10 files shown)


crates/mooncake/src/pkg/remove.rs
```diff
@@ -44,7 +43,6 @@ pub fn remove(
     module_dir: &Path,
     username: &str,
     pkgname: &str,
-    _registry_config: &RegistryConfig,
```
peter-jerry-ye, line 47: Notice that it's never used. This PR simply removes it and waits for future proper implementation

PR #1552, opened 2026-03-23
Skip incompatible test-only imports during moon check

peter-jerry-ye:
## Summary
- retain only backend-compatible auto-added test check roots during `moon check`
- keep source checking strict while warning when blackbox or whitebox checks are skipped
- add a planner regression fixture covering test-only `test-import` and `wbtest-import` target restrictions

## Testing
- cargo test -p moon target_backend_planning -- --nocapture
- cargo test -p moon backend_selection -- --nocapture


peter-jerry-ye:
Split to (with the same concept) #1564 and  #1567

PR #1563, opened 2026-03-25
improve expect test handling

Guest0x0:
In the future, the representation of `SourceLoc` may change from package + package-relative file name to module + module-relative file name. This PR make the expect test handling logic of `moon` ready for this future change, and perform some code simplification & cleanup.

Previously, expect test resolve the absolute path of tests by:

1. read the package and package-relative filename,
2. resolve the source directory of the package
3. append the package-relative filename to the package's source directory

This PR mainly modifies step (1) here: for expect test, the package part of source location can only be the very package being tested. So there is no need to receive it from the MoonBit side, as `moon` know which package is being tested.

After this simplification, `moon` no longer rely on package/module information from the MoonBit side. Next, this PR leaves only the base name part of the package-relative filename. This way, current code will work automatically when package-relative filename migrates to module-relative filename.

### DIFF
--- a/crates/moonbuild/src/expect.rs
+++ b/crates/moonbuild/src/expect.rs
@@ -28,7 +28,7 @@ use similar::TextDiff;
 use std::collections::BTreeSet;
 use std::collections::HashMap;
 use std::io::Write;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 pub trait PackageSrcResolver {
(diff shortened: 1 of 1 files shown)


crates/moonbuild/src/expect.rs
```diff
@@ -146,45 +146,26 @@ impl std::str::FromStr for ExpectFailedRaw {
...
 pub struct SnapshotResult {
-    pub loc: LocationJson,
-    pub args_loc: Vec<Option<LocationJson>>,
-    pub expect_file: PathBuf,
-    pub expect_content: Option<String>,
-    pub actual: String,
-    pub succ: bool,
+    loc: Location,
+    expect_file: PathBuf,
+    actual: String,
```
Guest0x0, line 168: After removing the unused `pub`, it turns out that many fields here are simply dead code

PR #1613, opened 2026-04-10
fixes #1612

### LINKED ISSUE #1612: Bug: incorrect minutes conversion in bench output
## Description

`auto_select_unit` in `crates/moonbuild/src/benchmark.rs` uses `us / 6e10` to convert microseconds to minutes, but `6e10` µs = 60,000 seconds = 1,000 minutes. The correct divisor is `6e7` (60 × 1e6).

As a result, any benchmark duration that reaches the minutes branch displays a value 1000× too small.

Additionally, the seconds-to-minutes cutoff is at `1e9` µs (1,000 seconds). A cutoff of `6e7` µs (60 seconds) would be more natural — otherwise durations like 999 seconds display as `"999.00 s"` instead of minutes.

## Steps to reproduce

```rust
// current behavior
auto_select_unit(60_000_001.0) // expects "  1.00 min", actually returns " 60.00 s"
auto_select_unit(1_000_000_001.0) // returns "  0.02 min" (should be " 16.67 min")
```

## Expected behavior

```rust
auto_select_unit(60_000_001.0)    // "  1.00 min"
auto_select_unit(1_000_000_001.0) // " 16.67 min"
```

Found while working on #1588.

bikallem:
Fixes: https://github.com/moonbitlang/moon/issues/1612

### DIFF
--- a/crates/moonbuild/src/benchmark.rs
+++ b/crates/moonbuild/src/benchmark.rs
@@ -50,10 +50,10 @@ fn auto_select_unit(us: f64) -> String {
         format!("{us:>6.2} µs")
     } else if us < 1e6 {
         format!("{:>6.2} ms", us / 1e3)
-    } else if us < 1e9 {
+    } else if us < 6e7 {
         format!("{:>6.2} s", us / 1e6)
     } else {
-        format!("{:>6.2} min", us / 6e10)
+        format!("{:>6.2} min", us / 6e7)
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/moonbuild/src/benchmark.rs
```diff
@@ -98,3 +98,13 @@ pub fn render_batch_bench_summary(msg: &str) {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn auto_select_unit_minutes() {
+        assert_eq!(auto_select_unit(60_000_001.0), "  1.00 min")
+    }
```
peter-jerry-ye: It would be great if you could add the other regression that you mentioned

bikallem: done.

bikallem:
@peter-jerry-ye I have added the additional test as requested.

peter-jerry-ye:
Thanks a lot!

PR #1617, opened 2026-04-15
feat: support prelude proof

Yu-zh:
Pass the path to the prelude proof to `moonc prove` and also checks the availability of the prelude directory.


crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
```diff
@@ -54,6 +54,8 @@ use crate::{
 
 use super::{BuildCommand, Commandline, compiler, context::BuildPlanLowerContext};
 
+const PRELUDE_PROOF_DIR: &str = "prelude_proof";
```
peter-jerry-ye: Constants belong to `crates/moonutil/src/common.rs`

PR #1624, opened 2026-04-17
fix(install): atomically replace installed binaries

### LINKED ISSUE #1850: moon install: reinstalling a native binary in place → macOS "Killed: 9" (SIGKILL) from stale code-signing cache
## Summary

On macOS (arm64), running `moon install` **again** for a native `is-main` package — overwriting a binary that already exists in `~/.moon/bin` and has been executed before — produces a binary that the kernel **SIGKILLs at exec** (`Killed: 9`, exit 137, no output), *before* `main` runs. The on-disk code signature is still reported *valid*; the kill is the macOS code-signing/AMFI cache rejecting the in-place-replaced binary. A fresh install (or re-sign / fresh inode) runs fine.

This bit a user of a MoonBit CLI: after `moon install`, every invocation (`--help`, subcommands, etc.) was killed with no output.

## Environment

- `moon 0.1.20260624 (99b20c1 2026-06-24)`, `moonc v0.10.1+1f52b86e1-nightly`
- macOS 26.4.1 (build 25E253), Apple Silicon (arm64)
- Binary: `Mach-O 64-bit executable arm64`, `Signature=adhoc` / `flags=0x20002(adhoc,linker-signed)` (the normal state for a moon-built arm64 binary)

## Symptom

```
$ openseek --help          # (a moon-installed native is-main binary)
zsh: killed     openseek --help
$ echo $?
137                         # 128 + 9 = SIGKILL, and zero stdout/stderr
```

Every invocation is killed identically — the process never reaches `main` (hence no output), so it is not a bug in the program.

## Evidence that it's the code-signing cache, not the binary

```
# The on-disk signature is VALID, yet the binary is killed when run:
$ codesign -v --verbose=2 ~/.moon/bin/openseek
~/.moon/bin/openseek: valid on disk
~/.moon/bin/openseek: satisfies its Designated Requirement
$ ~/.moon/bin/openseek --help ; echo $?      # 137 (killed)

# Re-signing the SAME file in place fixes it immediately:
$ codesign -s - --force ~/.moon/bin/openseek
$ ~/.moon/bin/openseek --help ; echo $?      # 0, runs correctly

# A plain copy of the (killed) installed binary to a NEW path also runs fine
# after an ad-hoc re-sign — i.e. the bytes/logic are correct.
```

So the file is a correct, validly-signed binary; the kernel kills it only at the path it was reinstalled over.

## Root cause (analysis)

`moon install` appears to install by writing/overwriting the binary **in place** at `~/.moon/bin/<name>` (same vnode/inode). macOS caches a code-signing decision (CDHash) tied to the file; when a previously-executed binary is overwritten in place with **different content**, the cached hash no longer matches the new content and AMFI kills the process at exec (`SIGKILL`) rather than re-validating. This is a well-known macOS gotcha for tools that replace an executable in place (e.g. `cp` over a running/previously-run binary). It only reproduces when a binary already existed at that path *and had been run*; a first-time install is fine.

## Reproduction

1. `moon install` a native `is-main` package on macOS arm64; run the installed binary (works — caches the signature).
2. Change the source, rebuild, and `moon install` **again** (overwrites `~/.moon/bin/<name>` in place with different content).
3. Run the installed binary → `Killed: 9` / exit 137, no output.
4. `codesign -s - --force ~/.moon/bin/<name>` (or `rm` it before step 2) → runs again.

## Suggested fix

Make `moon install` not overwrite the executable in place. Any of:

- **Atomic install via rename**: write to a temp file in the same dir, then `rename(2)` over the destination (new inode ⇒ no stale cache). This is also crash-safe.
- **Re-sign after copy**: `codesign -s - --force <dest>` on macOS post-install (ad-hoc is fine; it refreshes the cached CDHash).
- At minimum, `unlink` the destination before writing the new file.

Atomic rename is the cleanest and also avoids partially-written binaries.

## Workaround (for users hitting this today)

```bash
codesign -s - --force ~/.moon/bin/<name>     # or:  rm ~/.moon/bin/<name> && moon install
```

tonyfettes:
## Summary

This PR makes `moon install --bin <dir>` replace installed binaries atomically instead of overwriting the destination file in place.

The install command now treats `<dir>/<binary>` as the managed command entry:

- build output under `_build` is preserved
- the executable is copied into a temporary file inside the install directory
- the temporary file is persisted over `<dir>/<binary>` with an atomic same-directory replacement
- an existing destination symlink is replaced as the command entry instead of being followed

## Why

The previous `std::fs::copy(src, dst)` path opens an existing destination and truncates it before copying the new bytes. On macOS, reinstalling a native executable this way can poison the code-signing cache for the existing inode. The install succeeds, but later execution can be killed by the kernel before `main` runs.

Atomic replacement avoids mutating the old executable inode. Existing running processes keep the old file, and new launches resolve the installed path to the replacement file.

## Why this shape

The helper takes the install directory directly, stages the temporary file there, and returns the final installed path. That keeps the same-filesystem replacement invariant explicit without recomputing the parent from the destination path.

The change is intentionally scoped to binary installation: it defines the install operation as replacing the managed command entry while preserving the build artifact that the build graph owns.

### DIFF
--- a/crates/moon/src/cli/install_binary.rs
+++ b/crates/moon/src/cli/install_binary.rs
@@ -32,7 +32,6 @@ use moonutil::{
 };
 use semver::Version;
 use std::path::{Path, PathBuf};
-#[cfg(target_os = "macos")]
 use tracing::debug;
 
 use crate::{
@@ -672,19 +671,7 @@ fn build_and_install_packages(
         } else {
             pkg.binary_name.clone()
         };
-        let binary_dst = install_dir.join(dst_name);
-
-        #[cfg(target_os = "macos")]
-        macos_install_file(&binary_src, &binary_dst)?;
-
-        #[cfg(not(target_os = "macos"))]
-        std::fs::copy(&binary_src, &binary_dst).with_context(|| {
-            format!(
-                "Failed to copy binary from `{}` to `{}`",
-                binary_src.display(),
-                binary_dst.display()
-            )
-        })?;
+        let binary_dst = install_file_atomically(&binary_src, install_dir, &dst_name)?;
 
         #[cfg(unix)]
         {
@@ -713,30 +700,25 @@ fn build_and_install_packages(
     Ok(0)
 }
 
-/// macOS-only: copy `src` onto `dst` by writing to a sibling tempfile
-/// and atomically renaming it into place.
+/// Copy `src` into `install_dir/dst_name` by writing to a sibling tempfile
+/// and atomically replacing the destination command entry.
 ///
-/// Overwriting the destination with `fs::copy` would truncate it via
-/// `O_TRUNC`; on macOS the kernel SIGKILLs any process that re-execs the
-/// modified inode because the code-signature cache no longer matches the
-/// on-disk bytes. Renaming a fresh file into place avoids this — the
-/// old inode is unlinked but remains alive for the running process, and
-/// new executions resolve to the new inode.
+/// `moon install --bin <dir>` manages `<dir>/<binary>`, so an existing
+/// destination path is replaced rather than opened and overwritten in place.
+/// This avoids exposing a partially written executable and avoids platform
+/// traps such as macOS code-signature cache kills after truncating a Mach-O
+/// binary that has already been executed.
 ///
-/// The tempfile is created *inside* `install_dir` (a sibling of `dst`)
-/// so the final rename is same-filesystem and therefore atomic. `src`
-/// is copied rather than moved because it is the build graph's declared
-/// artifact and must remain in `_build` for subsequent no-op commands.
-///
-/// Linux has a related issue (`ETXTBSY` on truncate of a running text
-/// segment) and Windows has a sharing-violation failure mode, but
-/// those surface as loud errors the user can diagnose. Only the macOS
-/// SIGKILL is silent enough to justify the extra complexity here.
-#[cfg(target_os = "macos")]
-fn macos_install_file(src: &Path, dst: &Path) -> anyhow::Result<()> {
-    let install_dir = dst.parent().ok_or_else(|| {
-        anyhow::anyhow!("Install path `{}` has no parent directory", dst.display())
-    })?;
+/// The tempfile is created inside `install_dir` so the final persist is a
+/// same-filesystem replacement. `src` is copied rather than moved because it
+/// is the build graph's declared artifact and must remain in `_build` for
+/// subsequent no-op commands.
+fn install_file_atomically(
+    src: &Path,
+    install_dir: &Path,
+    dst_name: &str,
+) -> anyhow::Result<PathBuf> {
+    let dst = install_dir.join(dst_name);
 
     let tmp_path = tempfile::NamedTempFile::new_in(install_dir)
         .with_context(|| {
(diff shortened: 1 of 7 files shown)


peter-jerry-ye:
@codex Review

crates/moon/src/cli/install_binary.rs
```diff
@@ -645,6 +632,337 @@ fn build_and_install_packages(
...
+        let mut buffer = vec![0u8; total_size];
+
+        let flags = FILE_RENAME_FLAG_POSIX_SEMANTICS | FILE_RENAME_FLAG_REPLACE_IF_EXISTS;
+        buffer[0..4].copy_from_slice(&flags.to_le_bytes());
+        // RootDirectory stays zero (null).
+        let name_len_offset = HEADER_SIZE - 4;
+        buffer[name_len_offset..HEADER_SIZE].copy_from_slice(&(name_byte_len as u32).to_le_bytes());
+        for (i, &wc) in dst_wide.iter().enumerate() {
+            buffer[HEADER_SIZE + i * 2..HEADER_SIZE + i * 2 + 2].copy_from_slice(&wc.to_le_bytes());
+        }
```
peter-jerry-ye: Please use Rust struct to define it properly

crates/moon/src/cli/install_binary.rs
```diff
@@ -645,6 +632,337 @@ fn build_and_install_packages(
...
+    /// next install sweeps again.
+    pub fn sweep_stale_backups(dst: &Path) {
+        let Some(parent) = dst.parent() else { return };
+        let Some(dst_name) = dst.file_name() else {
+            return;
+        };
+
+        let mut prefix = dst_name.to_owned();
+        prefix.push(".old-");
+        let prefix_bytes = prefix.as_encoded_bytes();
```
peter-jerry-ye: Useless `as_encoded_bytes`?

crates/moon/src/cli/install_binary.rs
```diff
@@ -645,6 +632,337 @@ fn build_and_install_packages(
...
+                    path = %path.display(),
+                    err = %e,
+                    code = ?e.raw_os_error(),
+                    "could not sweep stale backup"
+                ),
+            }
+        }
+    }
+
+    fn backup_path_for(dst: &Path) -> PathBuf {
```
peter-jerry-ye: Why not just use `tempfile`?

tonyfettes:
Reviewer note: this branch is intentionally structured as a TDD stack:

1. RED: reproduce the macOS running-binary overwrite bug with the original `fs::copy` install path.
2. GREEN: switch the same-filesystem install path to `fs::rename`.
3. RED: add the cross-filesystem install regression test.
4. GREEN: handle cross-filesystem installs by falling back after `EXDEV`/cross-device rename failure.

The two RED commits are expected to fail when checked out individually; `HEAD` is the reviewed state. On `HEAD`, the install-relevant macOS stable/nightly jobs pass, as do Ubuntu stable/nightly and stable Windows. The remaining red checks are shared with other open PRs: Windows nightly `test_zombie_child_process` timeout and verification `moon_prove` snapshot drift.

crates/moon/src/cli/install_binary.rs
```diff
@@ -672,6 +678,57 @@ fn build_and_install_packages(
...
+        format!(
+            "Failed to copy binary from `{}` to `{}`",
+            src.display(),
+            tmp_path.display()
+        )
+    })?;
+    debug!(src = %src.display(), "staged via copy");
+
+    tmp_path
+        .persist(dst)
```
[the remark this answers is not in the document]
tonyfettes: Handled by making the macOS install path intentionally replace the destination command entry rather than follow it. The PR description now calls this out explicitly: `moon install --bin <dir>` manages `<dir>/<binary>`, so an existing destination symlink is replaced by the installed binary instead of writing through to its target.

PR #1686, opened 2026-05-17
moon: add profile mode for run via xctrace

Yu-zh:
## Summary
- add `moon run --profile` for macOS native profiling through `xcrun xctrace`
- force a standalone native executable for profiling and post-process xctrace time-profile exports with MoonBit demangling
- print a brief terminal report plus paths to the trace, text report, JSON report, and target stdout
- cover xctrace stack/backtrace export shapes and profile dry-run command output

## Tests
- `cargo clippy -p moon --all-targets -- -D warnings`
- `cargo test -p moon profile::tests --no-fail-fast`
- `cargo test -p moon test_moon_run_profile_dry_run_prints_xctrace_commands --no-fail-fast`

### DIFF
--- a/crates/moon/src/cli/profile.rs
+++ b/crates/moon/src/cli/profile.rs
@@ -142,7 +142,11 @@ fn run_profile_materialized(
         cli,
         &run_cmd,
         BuildRunExecutableOptions {
+            // Profiling needs a stable executable path for xctrace to launch.
+            // The TCC fast path may run directly from generated C instead.
             try_tcc_run: false,
+            // The dry-run output should show the profiled invocation, not the
+            // plain executable command that `moon run` would normally print.
             print_dry_run_run_command: false,
         },
     )?;
@@ -223,6 +227,8 @@ fn profile_run_subcommand(cmd: RunSubcommand) -> anyhow::Result<RunSubcommand> {
     {
         bail!("`moon run --profile` currently supports only `--target native`");
     }
+    // Time Profiler records a native process. Build release-with-symbols by
+    // default so samples are useful without requiring extra flags from users.
     build_flags.target = vec![SurfaceTarget::Native];
     if !build_flags.debug && !build_flags.release {
         build_flags.release = true;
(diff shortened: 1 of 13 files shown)


crates/moon/src/cli/profile.rs
```diff
@@ -0,0 +1,776 @@
...
+        if state != "Running" {
+            continue;
+        }
+
+        let weight = first_tag(row, "weight")
+            .and_then(|tag| resolve_value(tag, &weights, ""))
+            .unwrap_or(parsed.sample_weight_ms);
+
+        let Some(stack) = stack else { continue };
+        let stack = stack
```
peter-jerry-ye, line 474: One concern: `top_self` is computed after filtering frames through `is_profile_symbol`. If the real leaf frame is a system/runtime frame and the first retained frame is a MoonBit caller, this sample gets counted as MoonBit self time even though it is more like attributed/inclusive time.

Could we either only count self time when the original leaf frame is retained, or rename this metric so users don’t read it as precise leaf/self time?

crates/moon/src/cli/profile.rs
```diff
@@ -0,0 +1,776 @@
...
+    inclusive_counts: HashMap<String, usize>,
+    inclusive_ms: HashMap<String, f64>,
+}
+
+pub(crate) fn run_profiled_run(cli: &UniversalFlags, cmd: RunSubcommand) -> anyhow::Result<i32> {
+    if !cfg!(target_os = "macos") {
+        bail!("`moon run --profile` currently supports macOS only");
+    }
+
+    ensure_xctrace_available()?;
```
peter-jerry-ye, line 77: Small UX issue: `--dry-run` still requires `xcrun xctrace version`, and this check also runs before target validation. So on macOS without xctrace, `moon run --profile --target js --dry-run` may fail with an xctrace setup error instead of explaining that profile only supports native.

Could dry-run skip the xctrace availability check, or move target validation before probing xctrace?

peter-jerry-ye (commented):
Two small non-blocking review notes.

PR #1692, opened 2026-05-19
Refactor test invocation collection

Yu-zh:
## Summary
- extract collection of concrete test executable invocations from the test runner
- reuse the collected invocation list in normal sequential and parallel test execution
- keep build-only artifact collection on the same invocation path

## Tests
- cargo fmt --all
- cargo test -p moon js_test_build_only --no-fail-fast
- cargo test -p moon test_async_test_inline --no-fail-fast
- cargo clippy -p moon --all-targets -- -D warnings


Yu-zh:
@codex review

PR #1701, opened 2026-05-20
[codex] Delay native C toolchain resolution

peter-jerry-ye:
## Summary

- Replace infallible global native C compiler detection with cached fallible system/internal-tcc detection.
- Carry a native toolchain selection policy through RR planning instead of resolving a default compiler during compile config construction.
- Resolve the default native toolchain only when native work actually needs it and no package-level `cc`/`stub-cc` override exists.
- Remove test support's infallible `CC::default()` lookup.

## Why

`moon check --target wasm-gc` could initialize native C compiler detection through `CC::default()` even though the selected backend does not need a C compiler. If system compilers and bundled internal `tcc` were unavailable, this could fail unrelated non-native workflows.

The new flow keeps non-native builds structurally free of native toolchain selection while preserving native defaults:

- `MOON_CC` takes precedence.
- Debug native defaults prefer internal `tcc`, then system compilers.
- Release native defaults prefer system compilers, then internal `tcc`.
- Package `cc`/`stub-cc` overrides avoid probing the default path entirely.

## Validation

- `cargo fmt --check`
- `cargo check -p moonutil -p moonbuild-rupes-recta -p moon`
- `cargo clippy -p moonutil -p moonbuild-rupes-recta -p moon --all-targets -- -D warnings`


peter-jerry-ye:
Superseded by the split stack:

- #1709: delay native C toolchain resolution
- #1710: clean up native toolchain selection
- #1711: separate native execution model cleanup

PR #1720, opened 2026-05-27
fix(fmt): reject moon.mod migration with local deps

### LINKED ISSUE #1716: `rr_moon_mod` migration drops path-deps with no opt-out
`moon 0.1.20260522` enables `rr_moon_mod`. `moon fmt` then rewrites `moon.mod.json` to TOML `moon.mod`. The rewrite drops the `path` field from every dep. The new `import {}` block has no path-dep syntax. `moon check` fails on the migrated file. I couldn't find an opt-out.

## Actual behavior

`moon fmt` rewrites:

```json
{ "deps": { "dowdiness/loom": { "path": "./loom/loom", "version": "0.1.0" } } }
```

to:

```
import { "dowdiness/loom", }
```

`moon check` then fails:

```
moon.mod only supports versioned registry dependencies in `import`, found `dowdiness/loom`
```

A versioned form (`"dowdiness/loom@0.1.0"`) only resolves if the target is also listed in `moon.work` `members`. The migration tool doesn't promote path-deps into `moon.work`. I also tried inline-table forms on `import {}` (`{ path = "..." }`, `path "..."`) and the lexer rejected all of them.

**Cross-repo path-deps have no fix.** My `loom/loom/moon.mod.json` reaches into canopy's `lib/` from inside a git submodule:

```json
"dowdiness/text_change": { "path": "../../lib/text-change" }
```

`moon.work` can't list a path owned by another repo's workspace, so workspace-promotion doesn't help.

**No opt-out.** None of these disabled `rr_moon_mod`:

```sh
moon -Z=!rr_moon_mod fmt          # Unknown feature
moon -Z=-rr_moon_mod fmt          # Unknown feature
MOON_UNSTABLE="" moon version --all       # still enabled
MOON_UNSTABLE="wasi_link" moon version --all   # still enabled
```

The CDN at `cli.moonbitlang.com/binaries/<version>/...` only serves `latest` and `nightly`. Every versioned URL I tried — `0.1.20260522`, `0.1.20260512`, `stable`, `bleeding` — returns HTTP 403. Pinning to the last working release is not an option.

## Expected Behavior

Add a TOML syntax for path-deps in `import`, e.g.

```
import {
  "dowdiness/loom@0.1.0" { path = "./loom/loom" },
}
```

The legacy `moon.mod.json` loader already understands path-deps. This unblocks both same-repo and cross-repo path-deps.

The migration tool should preserve `path` rather than drop it silently. Until the syntax lands, `rr_moon_mod` needs an opt-out. Default it to off, or expose a `--unstable-feature=no-rr_moon_mod` / `MOON_UNSTABLE` switch. Archiving older CDN builds would also unblock pinning as a temporary workaround.

## Steps to Reproduce

Triggers every time on any module with a path-dep:

```sh
mkdir repro && cd repro
moon work init
# Writing path-deps including moon.mod.json
cat > moon.mod.json <<'EOF'
{
  "name": "test/proj",
  "version": "0.1.0",
  "deps": {
    "moonbitlang/x": { "path": "../x", "version": "0.4.0" }
  }
}
EOF
moon fmt        # drops the path field
moon check      # fails on the resulting import
```

## Environment

Operating System: Linux (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2)

```
moon 0.1.20260522 (84aa893 2026-05-22) ~/.moon/bin/moon
moonc v0.9.3+3d4544a9e (2026-05-22) ~/.moon/bin/moonc
moonrun 0.1.20260522 (84aa893 2026-05-22) ~/.moon/bin/moonrun

Feature flags enabled: rr_moon_mod,rr_moon_pkg
```

This is my working [repo](https://github.com/dowdiness/canopy).

## Checklist

- [x] (Optional) My case is minimal enough to be reproducible.

Yoorkin:
- Related issues: None
- PR kind: Bugfix

## Summary

Reject `moon.mod.json` to `moon.mod` migration when the module declares local dependencies, since `moon.mod` does not support them yet. The formatter now reports the existing `moon.work` guidance instead of generating an unsupported `moon.mod`.

Also adds regression coverage for both normal `fmt` and `fmt --dry-run`.

## Metadata

@myfreess

### DIFF
--- a/crates/moon/tests/test_cases/fmt_moon_mod/mod.rs
+++ b/crates/moon/tests/test_cases/fmt_moon_mod/mod.rs
@@ -311,6 +311,41 @@ fn test_reading_module_warns_when_moon_mod_shadows_json() {
 }
 
 #[test]
+fn test_fmt_moon_mod_json_migration_local_deps_fail() {
+    let dir = TestDir::new_empty();
+    std::fs::create_dir_all(dir.join("main")).unwrap();
+    std::fs::write(
+        dir.join("moon.mod.json"),
+        r#"{
+  "name": "test/local_deps",
+  "deps": {
+    "example/local": { "path": "../local" }
+  }
+}
+"#,
+    )
+    .unwrap();
+    std::fs::write(dir.join("main/moon.pkg"), r#"options("is-main": true)"#).unwrap();
+    std::fs::write(dir.join("main/main.mbt"), "fn main { println(1) }\n").unwrap();
+
+    let stderr = get_err_stderr(&dir, ["fmt", "--dry-run"]);
+    assert!(
+        stderr.contains("moon.mod does not support local dependency `example/local`"),
+        "{stderr}"
+    );
+    assert!(stderr.contains("moon.work"), "{stderr}");
+
+    let stderr = get_err_stderr(&dir, ["fmt"]);
+    assert!(
+        stderr.contains("moon.mod does not support local dependency `example/local`"),
+        "{stderr}"
+    );
+    assert!(stderr.contains("moon.work"), "{stderr}");
+    assert!(!dir.join("moon.mod").exists());
+    assert!(dir.join("moon.mod.json").exists());
+}
+
+#[test]
 fn test_fmt_moon_mod_local_deps_fail() {
     let dir = TestDir::new_empty();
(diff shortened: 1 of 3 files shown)


Yu-zh:
@codex review

PR #1725, opened 2026-06-01
[codex] Support LLVM test driver cfg

peter-jerry-ye:
## Summary

- Treat the LLVM backend like native in the generated MoonBit test driver template.
- Add an LLVM regression test for moon test --target llvm using the existing moon_test/succ fixture.

## Root Cause

The runtime already executes LLVM test binaries through the native-style CLI path, and MoonBuild passes --target llvm through to the generated driver compilation. The generated driver template only exposed the native entry point and CLI argument parser under #cfg(target="native"), so LLVM test drivers could compile without a #test_entry main.

## Validation

- moon fmt --check --manifest-path crates/moonbuild/template/test_driver_project/moon.mod
- cargo test -p moon --test mod test_moon_test_succ -- --nocapture

Note: the new regression test skips when the active toolchain does not provide LLVM stdlib artifacts, since LLVM is nightly-only.

### DIFF
--- a/crates/moon/tests/test_cases/moon_test/mod.rs
+++ b/crates/moon/tests/test_cases/moon_test/mod.rs
@@ -28,6 +28,7 @@ fn test_moon_test_succ() {
 }
 
 #[test]
+#[cfg(not(windows))]
 fn test_moon_test_succ_llvm() {
     let dir = TestDir::new("moon_test/succ");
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
The new native backend error is known and will fix it later.

PR #1750, opened 2026-06-04
fix: give an actionable error when moonbitlang/core is not bundled

### LINKED ISSUE #1718: Windows: Cannot inject the standard library moonbitlang/core after fresh install (v0.1.20260522)
# Bug Report: Windows - Cannot inject the standard library `moonbitlang/core`

## Summary
On Windows, MoonBit CLI v0.1.20260522 fails to load the standard library `moonbitlang/core` for any project, including freshly created ones via `moon new`. The error occurs across all build targets (js, wasm, wasm-gc, native).

## Version

```
moon 0.1.20260522 (84aa893 2026-05-22)
Feature flags enabled: rr_moon_mod,rr_moon_pkg
```

## Environment

- OS: Windows 11 (PowerShell / cmd)
- MoonBit installed via: `irm https://cli.moonbitlang.cn/install/powershell.ps1 | iex`

## Reproduction Steps

1. Fresh install MoonBit on Windows
2. Run `moon new test_project`
3. `cd test_project`
4. Run `moon build --target js` (or `moon test --target js`)

## Expected Behavior

Build/test should succeed using the built-in standard library.

## Actual Behavior

```
Error: failed to run build for target Js

Caused by:
    0: Failed to resolve the module dependency graph
    1: Cannot inject the standard library `moonbitlang/core`: Cannot load the core file
```

## Additional Context

- The standard library source **does exist** at `~/.moon/core/` (cloned from https://github.com/moonbitlang/core, `moon.mod` present)
- Also tried placing it at `~/.mooncakes/moonbitlang/core/` — same error
- Reinstalling the toolchain via the PowerShell script does **not** resolve the issue
- The `.moon/lib/` directory was temporarily deleted during reinstall; manually restoring `runtime_core.c` and other files from the ZIP distribution did not help
- All targets fail with the identical error message
- The `mooncake` CLI does not have an `install` subcommand to manually fetch core

## Suspected Cause

The `rr_moon_mod` / `rr_moon_pkg` feature flags may have changed the standard library discovery path on Windows, causing `moonc` to fail locating/injecting `moonbitlang/core`.

mvanhorn:
- Related issues: Fixes #1718
- PR kind: Bugfix (diagnostics)

## Summary

`moon build` on a fresh install where `~/.moon/lib/core` exists but was never bundled now fails with an actionable message that names the core directory and the exact remediation (`moon -C <core_dir> bundle --target all`), instead of the generic `Cannot load the core file`. A missing core directory gets its own message pointing at reinstalling the toolchain or `$MOON_CORE_OVERRIDE`.

## Why this matters

#1718 hit this on Windows: the old error reads like a corruption/path bug and sends users down the wrong debugging path, while peter-jerry-ye's confirmed fix is just running the bundle command. The change is confined to the `.context(...)` on `read_module_desc_file_in_dir` in `inject_std` (`crates/mooncake/src/resolver.rs`); the underlying error stays in the source chain and the `CannotInjectCore` wrapper still prefixes the standard-library context.

Output after this change (exists-but-unbundled path):

```
Cannot inject the standard library `moonbitlang/core`: Cannot load the core file. The standard library at `/tmp/.../unbundled_core` does not appear to be bundled. Run `moon -C /tmp/.../unbundled_core bundle --target all` to bundle it
```

## Metadata

- [x] Tests added/updated for bug fixes or new features - N/A for this diagnostics-only change; both failure paths verified by running the built `moon` binary against a minimal project with `MOON_CORE_OVERRIDE` pointing at an empty dir and a nonexistent dir (outputs above)
- [x] Compatible with Windows/Linux/macOS - message formatting only, no platform-specific APIs; `cargo check -p mooncake` and `cargo fmt --check` pass

### DIFF
--- a/crates/mooncake/src/resolver.rs
+++ b/crates/mooncake/src/resolver.rs
@@ -266,8 +266,24 @@ pub(crate) fn resolve_with_default_env(
 /// to the resolve graph, and mark it as the standard library.
 fn inject_std(res: &mut ResolvedEnv) -> anyhow::Result<()> {
     let core_dir = moon_dir::core();
-    let loaded_core =
-        read_module_desc_file_in_dir(&core_dir).context("Cannot load the core file")?;
+    let loaded_core = read_module_desc_file_in_dir(&core_dir).with_context(|| {
+        if core_dir.exists() {
+            format!(
+                "Cannot load the core file. The standard library at `{}` \
+                 does not appear to be bundled. Run `moon -C {} bundle --target all` \
+                 to bundle it",
+                core_dir.display(),
+                core_dir.display()
+            )
+        } else {
+            format!(
+                "Cannot load the core file. The standard library directory `{}` \
+                 does not exist. Reinstall the MoonBit toolchain, or set \
+                 $MOON_CORE_OVERRIDE to a checkout of moonbitlang/core",
+                core_dir.display()
+            )
+        }
+    })?;
     let source = ModuleSource::from_stdlib(&loaded_core, &core_dir);
     let id = res.add_module(source, Arc::new(loaded_core));
     res.register_stdlib(id);


peter-jerry-ye:
Although I think the fix is fine, I don't think such situation should happen often enough to add a fix. Either it's because the user is trying to develop `moonbitlang/core` without reading [`CONTRIBUTING`](https://github.com/moonbitlang/core/blob/ace5c88416720bbed83b1c55f2ab93d2ed68a8c2/CONTRIBUTING.md?plain=1#L27) or there's a bug.

Such a fix would hide away such unexpected situation. I would need a stronger reason for this fix.

peter-jerry-ye:
Also I'm still waiting for the clarification for that issue because it claimed to be a "fresh installation"

mvanhorn:
That's fair - no pressure on this one. If the reporter's clarification shows it really was a fresh install hitting a missing core, this is ready; if it turns out to be a broken dev setup, happy to close it as not worth the surface area. I'll leave it to your call once they respond.

PR #1760, opened 2026-06-09
fix: make implicit target preferences capability-aware

peter-jerry-ye:
## Why

Workspace and module preferred targets are preferences, but package capability still decides whether a backend can be used. The previous behavior applied that policy unevenly across commands.

Workspace mode also does not yet have a direct command-line selector for a member module. Users often approximate that with shell-expanded directory spans, which can include non-package directories or packages that do not support the selected backend. Path selectors therefore need to remain tolerant and keep the supported packages instead of treating every skipped match as an error.

## Correctness

This PR makes implicit package target selection choose the first supported candidate in this order: workspace `preferred_target`, module `preferred-target`, `wasm-gc`, then the remaining backends in backend order. Explicit `--target` remains a hard request: Moon does not fall back to another backend for a package selected for that planned run.

Broad workspace scans drop unsupported packages and unsupported backend groups, including packages whose effective `supported_targets` set is empty. Path-span selectors keep packages that support the planned backend and skip unsupported or non-package matches; they still fail when the selection resolves to no supported package. `moon run` remains quiet for implicit fallback.

## Minimality

The change keeps target preference logic centralized in the package filtering helpers, preserves the existing tolerant path-span UX, and updates the affected command docs and regression tests.


peter-jerry-ye:
We still need to discuss this.

PR #1762, opened 2026-06-09
Remove unused core bundle helpers

Yu-zh:
## Summary

- remove unused `moonutil::moon_dir::core_bundle` and `core_core`
- drop the obsolete release-bundle assertion from `moon_dir` tests

## Compatibility check

- searched local `../mooncake` clone for `core_bundle` / `core_core`: no matches
- searched local `../mooncake` clone for `moon_dir::` uses: only `credentials_json()` is referenced

## Tests

- `cargo check -p moonutil`
- `cargo test -p moonutil test_moon_dir`

### DIFF
--- a/crates/moonutil/src/moon_dir.rs
+++ b/crates/moonutil/src/moon_dir.rs
@@ -19,8 +19,6 @@
 use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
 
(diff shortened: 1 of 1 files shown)


peter-jerry-ye:
The dependency should be reversed : the layout info should be centered to toolchain information, instead of letting the rr (build kernel) to rediscover again.

PR #1766, opened 2026-06-11
Unify implicit target preference policy

peter-jerry-ye:
## Why

Implicit target selection was split across two mental models: package-grouped commands used module preference before workspace preference, while shared build planning used workspace preference first. That made workspace preferred_target hard to reason about and made moon run differ from build/check/test behavior.

## What changed

- Use workspace preferred_target as the first execution policy candidate for implicit target selection, then module preferred_target, then wasm-gc, then the first remaining backend supported by the module.
- Resolve implicit fallback once per module; package-level supported_targets only filter packages within that module target.
- If a single explicitly selected package narrows below the module-selected backend, report the unsupported backend and require --target for that package-specific backend. Multi-path scans remain best-effort.
- Make moon run resolve the selected package before build planning so it can use the same module-aware fallback policy.
- Keep moon info canonical pkg.generated.mbti output package-local: module preferred-target, then wasm-gc.
- Update developer/manual docs and add regression coverage for workspace override, unsupported workspace-target fallback, no per-package fallback inside a module, and moon info canonical output.

## Why this is minimal

The shared package preference helper now owns module-level execution target fallback, moon run only changes where it obtains the implicit target, and moon info gets a local canonical-output rule instead of depending on execution target resolution.


peter-jerry-ye:
We are removing the workspace preferred-target directly.

PR #1773, opened 2026-06-11
Make runtime command construction fallible

peter-jerry-ye:
## Why

Runtime command construction still had direct panic paths around serializing test arguments and assembling commands. These are a natural next slice after mechanical unwrap cleanup because the command builder is a shared boundary used by `run`, `runwasm`, tests, and binary dependency installation.

## What Changed

This PR changes `crate::run::command_for` to return `anyhow::Result<Command>` and propagates failures through its callers. It also replaces adjacent command assembly unwraps in the binary dependency installer with contextual errors.

## Why This Is Correct

Serializing test arguments is required for correct test execution. Returning an error is preferable to either panicking or silently constructing a command without the intended arguments.

The change is limited to the command-construction boundary and direct callers, keeping broader metadata and build-graph invariant changes out of this PR.

### DIFF
--- a/crates/moon/src/cli/run.rs
+++ b/crates/moon/src/cli/run.rs
@@ -248,7 +248,7 @@ fn run_inline_source_as_single_file(
     let source = cmd
         .command
         .clone()
-        .expect("inline script should be present when `moon run -e` is selected");
+        .context("inline script should be present when `moon run -e` is selected")?;
 
     run_source_as_single_file(
         cli,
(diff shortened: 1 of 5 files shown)


peter-jerry-ye:
Closing this split: making command_for return Result mostly handled invariant failures, especially TestArgs serialization, and added API noise rather than improving user-facing error handling. We can keep documented expects or narrow lint allowances for those cases instead.

PR #1774, opened 2026-06-11
Report moon metadata failures as errors

peter-jerry-ye:
## Why

Some remaining panic paths represent malformed or missing build metadata discovered while running user-facing commands. These are still command failures, but routing them through the normal error path gives a clearer diagnostic than an internal panic.

## What Changed

This PR keeps only fail-fast conversions where the command still stops:

- package name matching reports unsupported multi-module metadata as an error
- MBTI output grouping reports unexpected artifact shape as an error
- binary dependency package selection reports unsupported local-module shape as an error
- test update and test artifact collection report missing artifacts/work entries as errors
- parallel test worker setup reports runtime or queue-lock failures through the test command result

## Why This Is Correct

The exceptional paths still terminate the command; this PR does not add fallback behavior, cancellation behavior, or log-and-continue recovery. The previous invariants are preserved, but failure now goes through the command error path with context instead of panicking.

This is intentionally separate from the mechanical unwrap cleanup in #1771. The fallible runtime command-builder split was closed because that API change mostly handled internal invariants rather than user-facing errors.

### DIFF
--- a/crates/moon/src/cli/build.rs
+++ b/crates/moon/src/cli/build.rs
@@ -523,7 +523,7 @@ fn resolve_selected_build_packages(
             resolve_output.local_modules(),
             package_filter,
             output,
-        );
+        )?;
         if let Some(target_backend) = target_backend {
             ensure_packages_support_backend(resolve_output, pkgs.iter().copied(), target_backend)?;
         }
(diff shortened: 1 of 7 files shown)


peter-jerry-ye:
Closing this split as well. We will focus on the concrete watch-path filesystem failures instead of broadly converting metadata invariants from panic to Result.

PR #1776, opened 2026-06-12
Unify target layout path resolution

peter-jerry-ye:
## Summary

This PR introduces a dedicated target layout module for Rupes Recta and routes target-directory artifact path computation through it.

It moves the old lowering-owned `PlannedArtifact` path mapping into `ArtifactPathResolver`, wires compile/lowering/build metadata/all-pkgs/fmt through the same resolver, and removes the old `LegacyLayout` module. It also keeps installed stdlib artifact paths owned by `moonutil::toolchain`, with the resolver only composing those selected toolchain facts where concrete artifact paths are needed.

## Why

Target output paths were computed in multiple places: the old layout module owned many filenames, lowering context owned logical artifact mapping, and some lowerers derived scratch/output directories from target paths directly. That made layout behavior harder to audit and made stdlib/toolchain routing look like part of target output layout.

## Why This Is Correct

`TargetLayout` now owns only paths under the selected target directory, while `ArtifactPathResolver` composes that layout with the already-selected stdlib path for callers that need concrete physical artifact paths. Metadata and `all_pkgs.json` reuse the same resolver as lowering, so they no longer rebuild layout facts separately. The native executable suffix behavior is preserved while removing the old `legacy_behavior` flag.

## Scope

The change is intentionally mechanical around path ownership: it preserves existing mono/workspace placement behavior, keeps toolchain stdlib routing separate, and updates the developer docs to match the new boundary.


peter-jerry-ye:
@codex Review

PR #1777, opened 2026-06-12
Fix moon install source build isolation

peter-jerry-ye:
## Why

`moon install` source builds were planned and executed from inside the installer, which made pre-build command working-directory semantics unclear and allowed caller workspace selection to affect source installs. This could break packages whose pre-build commands are written relative to the cloned checkout, downloaded module, or source workspace root.

## What

- Delegate source installation builds through normal `moon build` with an explicit source execution root.
- Preserve an applicable `moon.work` from the source being installed.
- Force workspace mode off when the source has no workspace, preventing caller workspace state from leaking into the install build.
- Resolve installable packages from the selected module instead of assuming the first resolved local module.
- Update prebuild and binary-install references to document the source install behavior.
- Add regression fixtures for local, git, and workspace source installs.

## Why this is correct

The installer still uses internal resolution for source acquisition, package selection, and artifact location, but the actual compilation path is the same `moon build` path users exercise directly. The delegated build receives either the source workspace manifest or `MOON_WORK=off`, so workspace behavior is explicit and source-scoped.

## Scope

This keeps the change local to binary install orchestration, prebuild argv0 path resolution, artifact lookup, docs, and focused regression tests.


peter-jerry-ye:
This problem is actually solved in #1781.

PR #1807, opened 2026-06-24
feat(config): allow formatter/source configurations in top-level

Yoorkin:
- Related issues: None
- PR kind: feature

## Summary

Support top-level DSL fields emitted by moonfmt for manifest migration:

- Allow `source = ...` as a top-level field in `moon.mod`
- Allow `formatter(...)` as a top-level field in `moon.pkg`
- Keep existing `options(formatter: ...)`，options(source: ...)  behavior compatible

## Metadata

### DIFF
--- a/crates/moonutil/src/common.rs
+++ b/crates/moonutil/src/common.rs
@@ -342,6 +342,7 @@ pub fn read_module_from_dsl(path: &Path) -> anyhow::Result<MoonMod> {
         ("license", false),
         ("keywords", false),
         ("description", false),
+        ("source", false),
         ("supported_targets", false),
         ("preferred_target", false),
     ]);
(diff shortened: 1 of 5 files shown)


peter-jerry-ye:
Do we still keep `source` field?

Yoorkin:
> Do we still keep `source` field?

Yes. We'll keep this because some projects still use it, but we don't recommend it.

peter-jerry-ye:
If we don't recommend it, why do we promote it to top-level?

Yoorkin:
> If we don't recommend it, why do we promote it to top-level?

The fields in `options(...)` are used to keep compatibility for legacy `.json` configurations, which need to be redesigned. Any stable configuration items (whether recommended or not) will move out of `options(...)` in the end.

PR #1813, opened 2026-06-25
Align async Wasm host offset ABI

peter-jerry-ye:
## Why

`moonbitlang/async` changed the directory entry path to decode names from a c-buffer using an entry-relative offset. The Wasm-support async branch also needs to sit on top of that upstream refactor, otherwise it still imports the old raw directory-name and no-offset string/c-buffer host shapes.

## What changed

- Rebase `moonbitlang/async` branch `codex/wasm-async-fs-host-upstream` onto async `main` and point the submodule at its new head.
- Keep `.gitmodules` tracking `codex/wasm-async-fs-host-upstream`, not `main`.
- Replace the raw directory name ABI with `fs/dir_entry_get_name_offset`.
- Align `c_buffer/blit_to_c` and `c_buffer/blit_from_c` with upstream's source/destination offset signatures.
- Keep the Rust `os_string` host adapter, but update it to decode `(buffer, offset, len)` instead of the old no-offset shape.
- Keep `errno_to_string` buffers shaped like native C strings so offset-aware `@os_string.decode` can use `len = -1` safely.

## Why this is correct

The async submodule branch remains the Wasm-support branch, now rebased over upstream's offset refactor. The Rust host exposes the ABI that the rebased Wasm branch imports: directory names are read by offset, c-buffer copies accept both source and destination offsets, and native string decoding remains host-owned for platform-specific UTF-8/UTF-16 handling.

## Scope

The change is limited to the Rust async host ABI adaptation, the rebased async submodule pointer/tracking metadata, and the existing moonrun async-host fixture.

### DIFF
--- a/crates/moonrun/src/async_api/c_buffer.rs
+++ b/crates/moonrun/src/async_api/c_buffer.rs
@@ -31,15 +31,18 @@ pub(super) fn is_null(_context: &mut ImportContext, ptr: u64) -> i32 {
 pub(super) fn blit_to_c(
     context: &mut ImportContext,
     dst: u64,
+    dst_offset: i32,
     src: i32,
-    offset: i32,
+    src_offset: i32,
     len: i32,
 ) -> AsyncHostResult<()> {
-    let src_len = checked_add_i32(offset, len)?;
+    let src_len = checked_add_i32(src_offset, len)?;
     let host = context.host;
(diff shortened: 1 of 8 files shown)


peter-jerry-ye:
Nightly is breaking due to compiler changes. Force merged.

PR #1822, opened 2026-06-26
support x86_64-pc-windows-msvc

Kaida-Amethyst:
支持x86_64-pc-windows-msvc目标

### DIFF
--- a/crates/moon/src/rr_build/mod.rs
+++ b/crates/moon/src/rr_build/mod.rs
@@ -44,7 +44,8 @@ use moonbuild_rupes_recta::{
     fmt::{FmtConfig, FmtResolveOutput},
     intent::UserIntent,
     model::{
-        Artifacts, BuildPlanNode, NativeTarget, PackageId, RunBackend, TargetKind, TccRunConfig,
+        Artifacts, BuildPlanNode, DirectNativeMode, NativeBackendMode, NativeTarget, PackageId,
+        RunBackend, TargetKind, TccRunConfig,
     },
     prebuild::{PrebuildEnvironment, run_prebuild_config},
     target_layout::{ArtifactPathResolver, TargetLayout},
@@ -341,26 +342,30 @@ impl CompilePreConfig {
             _ => None,
         };
         info!("New native target: {:?}", native_target);
-
-        let (run_backend, tcc_run) = match target_backend {
-            TargetBackend::Wasm => (RunBackend::Wasm, None),
-            TargetBackend::WasmGC => (RunBackend::WasmGC, None),
-            TargetBackend::Js => (RunBackend::Js, None),
-            TargetBackend::Native => (
-                RunBackend::Native,
-                if native_target.is_some() {
+        let (run_backend, native_mode) = match target_backend {
+            TargetBackend::Wasm => (RunBackend::Wasm, NativeBackendMode::GeneratedC),
+            TargetBackend::WasmGC => (RunBackend::WasmGC, NativeBackendMode::GeneratedC),
+            TargetBackend::Js => (RunBackend::Js, NativeBackendMode::GeneratedC),
+            TargetBackend::Native => {
+                let native_mode = if let Some(native_target) = native_target {
                     info!("Disabling `tcc -run`: new native backend selected");
-                    None
-                } else {
+                    NativeBackendMode::DirectObject(
+                        self.direct_native_mode(native_target, input_nodes)?,
+                    )
+                } else if let Some(tcc_run) =
                     self.select_tcc_run_config(resolve_output, input_nodes, output)
-                },
-            ),
-            TargetBackend::LLVM => (RunBackend::Llvm, None),
+                {
+                    NativeBackendMode::TccRun(tcc_run)
+                } else {
+                    NativeBackendMode::GeneratedC
+                };
+                (RunBackend::Native, native_mode)
+            }
+            TargetBackend::LLVM => (RunBackend::Llvm, NativeBackendMode::GeneratedC),
         };
         info!(
-            "Final run backend: {:?}, tcc-run enabled: {}",
-            run_backend,
-            tcc_run.is_some()
+            "Final run backend: {:?}, native mode: {:?}",
+            run_backend, native_mode
         );
         let stdlib_path = if std {
             Some(moonutil::toolchain::core())
@@ -378,8 +383,7 @@ impl CompilePreConfig {
         Ok(CompileConfig {
             target_dir: self.target_dir,
             target_backend: run_backend,
-            native_target,
-            tcc_run,
+            native_mode,
             opt_level: self.opt_level,
             action: self.action,
             debug_symbols: self.debug_symbols,
@@ -426,6 +430,37 @@ impl CompilePreConfig {
         info!("`tcc -run` availability: true");
         Some(TccRunConfig::new(internal_tcc))
     }
+
+    fn direct_native_mode(
+        &self,
+        native_target: NativeTarget,
+        input_nodes: &[BuildPlanNode],
+    ) -> anyhow::Result<DirectNativeMode> {
+        match native_target {
+            NativeTarget::X86_64PcWindowsMsvc if needs_native_c_toolchain(input_nodes) => {
+                let toolchain = compiler_flags::windows_msvc_native_toolchain(None).with_context(
+                    || "failed to resolve Windows MSVC native toolchain for direct native mode",
+                )?;
+                Ok(DirectNativeMode::resolved_windows_msvc(toolchain))
+            }
+            NativeTarget::X86_64PcWindowsMsvc => Ok(DirectNativeMode::pending_windows_msvc()),
+            NativeTarget::Aarch64AppleDarwin | NativeTarget::X86_64UnknownLinuxGnu => {
+                Ok(DirectNativeMode::generic(native_target))
+            }
+        }
+    }
+}
+
+fn needs_native_c_toolchain(input_nodes: &[BuildPlanNode]) -> bool {
+    input_nodes.iter().any(|node| {
+        matches!(
+            node,
+            BuildPlanNode::BuildCStub(_, _)
+                | BuildPlanNode::ArchiveOrLinkCStubs(_)
+                | BuildPlanNode::MakeExecutable(_)
+                | BuildPlanNode::BuildRuntimeLib
+        )
+    })
 }
 
 /// Read in the commandline flags and build flags to create a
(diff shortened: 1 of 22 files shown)


crates/moonbuild-rupes-recta/src/build_lower/compiler/msvc.rs
peter-jerry-ye, line 1: Use https://docs.rs/find-msvc-tools/latest/find_msvc_tools/ directly

peter-jerry-ye (commented):
Focused review comments from the RR design/correctness pass. The default `cl.exe` path looks directionally coherent, but I think the toolchain resolution boundary and accepted toolchain set need tightening before this is safe to merge.

crates/moonbuild-rupes-recta/src/build_plan/builders.rs
```diff
@@ -89,6 +89,9 @@ impl<'a> BuildPlanConstructor<'a> {
 
     fn effective_native_toolchain(&self, package_cc: Option<&CC>) -> anyhow::Result<Toolchain> {
         debug_assert!(self.build_env.target_backend.is_native());
+        if self.build_env.native_target == Some(NativeTarget::X86_64PcWindowsMsvc) {
+            return moonutil::compiler_flags::windows_msvc_native_toolchain(package_cc);
```
peter-jerry-ye: This resolves the Windows MSVC toolchain during build-plan construction. `windows_msvc_native_toolchain` is not a pure data lookup: it reads env/PATH/files, may run `vswhere`/`cmd.exe`/`vcvarsall`, and caches host state. That makes RR planning host-stateful. Can we resolve this before entering RR and pass explicit immutable toolchain/env data into the plan instead?

crates/moonutil/src/compiler_flags.rs
```diff
@@ -140,6 +164,259 @@ const CAN_USE_MOONBITRUN: bool = false;
...
+pub fn resolve_windows_msvc_toolchain() -> anyhow::Result<Toolchain> {
+    let environment = resolve_windows_msvc_environment()?;
+    let cl_exe = environment.cl_exe.display().to_string();
+    let cc = CC::try_from_path(&cl_exe)
+        .with_context(|| format!("failed to resolve MSVC compiler at {cl_exe}"))?;
+    Ok(Toolchain::from_path_probe(cc).with_msvc_environment(environment))
+}
+
+fn ensure_windows_msvc_compatible(cc: &CC) -> anyhow::Result<()> {
+    if cc.is_msvc() || cc.targets_msvc() {
```
peter-jerry-ye: This accepts any compiler whose dumped triple contains `msvc`, but the MSVC-specific lowering only takes the `cc.is_msvc()` path. A clang driver targeting `x86_64-pc-windows-msvc` can pass selection and then miss vcvars INCLUDE/LIB paths plus `WINDOWS_MSVC_DEFAULT_LIBS`; wrong-arch MSVC triples also pass. Please either narrow this to the exact implemented x86_64 `cl`/`clang-cl` support, or route clang-MSVC through complete MSVC env/default-lib handling.

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
```diff
@@ -934,6 +934,10 @@ impl<'a> LoweringContext<'a> {
             Some(&output_file.display().to_string()),
             self.opt.compiler_paths(),
         );
+        compiler::msvc::add_environment_include_paths(
```
peter-jerry-ye: Appending MSVC include paths after `make_cc_command_resolved` splits command ownership between the generic native command builder and RR lowering. That makes C-stub compile/link behavior easier to drift. Prefer carrying the include/lib env data into the native command builder, or precomputing it into explicit plan metadata and keeping argv expansion in one layer.

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
```diff
@@ -1277,21 +1281,32 @@ impl<'a> LoweringContext<'a> {
...
-            &self.opt.compiler_paths().lib_path,
-        );
-
-        let commandline = if run_dsymutil {
-            commandline_with_dsymutil(&linker_cmd, &dest)
-        } else {
-            linker_cmd.into()
-        };
+        let commandline =
+            if self.opt.native_target == Some(NativeTarget::X86_64PcWindowsMsvc) && cc.is_msvc() {
```
peter-jerry-ye: This new branch needs an integration-level lowered-command test or snapshot that covers the complete MSVC command shape: env `/LIBPATH` entries, propagated package link config, default Windows libs, and target-specific branch selection. The helper tests validate fragments, but they do not exercise this integration point.

peter-jerry-ye: Consider to add guard for this case since not using msvc is not acceptable for new native backend

crates/moonbuild-rupes-recta/src/build_lower/compiler/msvc.rs
Kaida-Amethyst: find-msvc的部分去除，命令拼装的逻辑保留了，之后windows如果用其它的c编译器特别是clang.exe之类的，需要在这里新增命令拼装逻辑

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
Kaida-Amethyst: Added an integration-level lowered-command test

Kaida-Amethyst: fixed

Kaida-Amethyst: added

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
Kaida-Amethyst: fixed

crates/moonbuild-rupes-recta/src/build_plan/builders.rs
Kaida-Amethyst: fixed

crates/moonutil/src/compiler_flags.rs
```diff
@@ -140,6 +163,156 @@ const CAN_USE_MOONBITRUN: bool = false;
...
+    }
+
+    resolve_windows_msvc_toolchain()
+}
+
+pub fn windows_msvc_toolchain_from_resolved(
+    resolved: &Toolchain,
+    package_cc: Option<&CC>,
+) -> anyhow::Result<Toolchain> {
+    if let Some(package_cc) = package_cc {
```
peter-jerry-ye: This path bypasses the existing `Toolchain::with_package_override` precedence rule. For the non-Windows native path, `MOON_CC` wins over a package `link.native.cc` override and only emits a warning; here a pre-resolved `MOON_CC=cl.exe` toolchain can still be replaced by the package CC, or even fail if that package CC is non-MSVC. Please preserve `ToolchainSource::EnvOverride` semantics before applying package-level overrides.

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
```diff
@@ -924,11 +925,11 @@ impl<'a> LoweringContext<'a> {
             .display()
             .to_string();
 
-        let cc = info.effective_native_toolchain.cc().clone();
-        let cc_cmd = make_cc_command_resolved(
-            cc,
+        let cc_cmd = make_cc_command_resolved_for_toolchain(
```
peter-jerry-ye, line 929: The C-stub object path does not encode the same MSVC runtime-library policy as the rest of the new Windows native path. `compile_runtime_command` explicitly emits `/MT` and final linking hard-codes `libcmt.lib`, while C stubs are still compiled by the generic helper with no explicit runtime option and no test that links a package with stubs. MSVC requires all modules in one link invocation to use the same runtime-library option, so this should either force the stub compile to `/MT` as well or centralize/test the runtime mode invariant for MSVC native builds.

crates/moon/src/rr_build/mod.rs
```diff
@@ -346,6 +346,14 @@ impl CompilePreConfig {
             _ => None,
         };
         info!("New native target: {:?}", native_target);
+        let native_toolchain = match native_target {
```
peter-jerry-ye: This resolves the Windows MSVC toolchain before the build plan knows whether the requested operation actually needs a native linker/compiler. For example, native check-style operations can build/check MoonBit artifacts without linking an executable, but they would now fail up front if MSVC Build Tools cannot be found. Consider deferring this resolution until the plan creates a runtime, C-stub, or executable action that needs the toolchain.

peter-jerry-ye (commented):
Reviewed the refreshed PR head after the upstream changes. The MSVC support is directionally much cleaner now, but I found two correctness blockers and one medium-risk early-failure issue.

peter-jerry-ye:
Suggestion on the native-mode shape: I think it would help to make the MSVC direct-native path a first-class mode instead of carrying it as separate optional fields such as `native_target`, `tcc_run`, and the resolved toolchain.

The existing lowering already has the right split conceptually: `RunBackend::Native` becomes a `CBackend`, and `CBackend` chooses between generated-C, direct-object native, and `tcc -run`. For MSVC, I’d make that direct-object branch carry its own resolved mode value, for example `DirectNativeMode::WindowsMsvc { toolchain }`, so the invariants are attached to one value:

- selected only for `x86_64-pc-windows-msvc`
- `tcc -run` is impossible in this mode
- effective compiler is `cl.exe` or `clang-cl.exe`
- runtime object, C stubs, and final link share one CRT policy, e.g. `/MT`
- toolchain discovery stays outside RR planning/lowering, preserving RR purity

That should make the support boundary explicit without creating a separate user-visible backend.

crates/moon/src/rr_build/mod.rs
Kaida-Amethyst: fixed

crates/moonbuild-rupes-recta/src/build_lower/lower_build.rs
Kaida-Amethyst: fixed

crates/moonutil/src/compiler_flags.rs
Kaida-Amethyst: fixed

Kaida-Amethyst:
> Suggestion on the native-mode shape: I think it would help to make the MSVC direct-native path a first-class mode instead of carrying it as separate optional fields such as `native_target`, `tcc_run`, and the resolved toolchain.
> 
> The existing lowering already has the right split conceptually: `RunBackend::Native` becomes a `CBackend`, and `CBackend` chooses between generated-C, direct-object native, and `tcc -run`. For MSVC, I’d make that direct-object branch carry its own resolved mode value, for example `DirectNativeMode::WindowsMsvc { toolchain }`, so the invariants are attached to one value:
> 
> * selected only for `x86_64-pc-windows-msvc`
> * `tcc -run` is impossible in this mode
> * effective compiler is `cl.exe` or `clang-cl.exe`
> * runtime object, C stubs, and final link share one CRT policy, e.g. `/MT`
> * toolchain discovery stays outside RR planning/lowering, preserving RR purity
> 
> That should make the support boundary explicit without creating a separate user-visible backend.

Refactored the native-mode shape

PR #1864, opened 2026-07-07
Remove regex from moonutil front matter parsing

peter-jerry-ye:
## Why

`moonutil` only used its direct `regex` dependency to find YAML front matter in `.mbt.md` files. That makes the shared utility crate pull in a heavier dependency for a simple marker-line parse.

## Correctness

The parser now scans the opening and closing front-matter marker lines directly, preserving the extracted YAML slice before handing it to the existing YAML parser. The added tests cover LF, CRLF, missing front matter, unclosed front matter, and an indented marker that should not terminate the block.

## Scope

This only removes `regex` from `moonutil` and updates the lockfile dependency edge. The workspace `regex` entry remains because other crates still use it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1799,7 +1799,6 @@ dependencies = [
  "nucleo-matcher",
  "petgraph",
  "pulldown-cmark",
- "regex",
  "schemars",
  "semver",
  "serde",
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
Closing this in favor of keeping `regex` for now and handling the `moonutil` work as an ownership/readability split in #1865.

PR #1866, opened 2026-07-07
fix: discover MSVC toolchain on Windows

peter-jerry-ye:
## Why

Windows generated-C native builds currently depend on `cl.exe` being available from the caller's environment. That means users often need to launch from a Visual Studio Developer Command Prompt even when the installed MSVC tools can be discovered programmatically.

## What Changed

- Discover the default Windows MSVC toolchain with `find-msvc-tools` when no explicit native C compiler is configured.
- Attach the discovered MSVC include and library paths to generated-C compile and link commands.
- Keep explicit `MOON_CC`, package `link.native.cc`, and independently resolved `stub_cc` behavior intact.
- Remove active CI setup of the MSVC developer command environment so Windows CI exercises automatic discovery.
- Clarify the generated-C native backend, direct object native target, native toolchain, ABI, and CRT terminology in developer docs.

## Why This Is Correct

The generated-C native backend only needs a usable C compiler/linker driver. On Windows, MSVC discovery now supplies the compiler path plus the include and library paths that were previously inherited from the developer environment. Explicit user compiler choices still take precedence, and incompatible stub/executable compiler combinations are left to fail naturally at link time.

## Scope

This does not introduce cross compilation or full `vcvarsall` environment emulation. The discovery target follows the current 64-bit Windows host architecture only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,10 +162,6 @@ jobs:
           "$env:USERPROFILE\.moon\bin" | Out-File -FilePath $env:GITHUB_PATH -Append
           "$env:GITHUB_WORKSPACE\target\debug" | Out-File -FilePath $env:GITHUB_PATH -Append
 
-      - name: Setup MSVC
-        if: ${{ matrix.os == 'windows-latest' }}
-        uses: ilammy/msvc-dev-cmd@v1
-
       - name: Build
         run: cargo build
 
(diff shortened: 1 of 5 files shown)


peter-jerry-ye:
Closing in favor of the current n2 cwd/env based MSVC support work, which preserves the full find-msvc-tools command environment instead of translating it into partial /I and /LIBPATH flags.

PR #1869, opened 2026-07-07
Escape Windows solver paths in why3 config

kokic:
## Summary

This PR escapes backslashes in Windows solver executable paths before writing them to the generated why3.conf. Why3 config strings treat backslashes as escape characters, so native Windows paths like C:\Users\...\z3.exe could be parsed incorrectly and cause prover invocation failures. The change preserves the original Windows path spelling, documents the reasoning, and adds a regression test for generated solver paths.

### DIFF
--- a/crates/moon/src/cli/prove.rs
+++ b/crates/moon/src/cli/prove.rs
@@ -477,9 +477,16 @@ fn try_detect_solver(spec: &SolverSpec) -> Option<DetectedSolver> {
 
     Some(DetectedSolver {
         name: spec.name,
-        path: path_str,
+        path: escape_windows_path(&path_str),
         version: version.to_string(),
     })
+}
+
+fn escape_windows_path(s: &str) -> String {
+    // Why3 configuration strings treat backslashes as escapes. Keep the native
+    // Windows path spelling instead of normalizing to `/`, so the generated
+    // prover path stays the exact executable path returned by PATH/env lookup.
+    s.replace('\\', "\\\\")
 }
 
 fn generate_why3_config(solvers: &[DetectedSolver]) -> String {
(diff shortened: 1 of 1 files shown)


crates/moon/src/cli/prove.rs
```diff
@@ -481,11 +481,18 @@ fn try_detect_solver(spec: &SolverSpec) -> Option<DetectedSolver> {
 
     Some(DetectedSolver {
         name: spec.name,
-        path: path_str,
+        path: escape_windows_path(&path_str),
```
peter-jerry-ye: This fixes the generated config for the current flow, but it moves Why3 config escaping into solver detection. Could we keep `DetectedSolver.path` as the raw executable path and escape at the `generate_why3_config` write boundary instead? Otherwise this field looks safe to reuse as a real path, but on Windows it now contains config-escaped backslashes.

crates/moon/src/cli/prove.rs
```diff
@@ -911,6 +918,14 @@ mod tests {
         assert!(!config.contains("Alt-Ergo"));
     }
 
+    #[test]
+    fn test_generate_config_escapes_windows_solver_path() {
+        let path = escape_windows_path(r"C:\Users\guest\bin\z3.exe");
```
peter-jerry-ye: This test pre-escapes the fixture before calling `generate_why3_config`, so it would still pass if the production change were removed. Could we pass the raw Windows path into `make_solver` and assert that `generate_why3_config` emits escaped backslashes? That would test the actual regression boundary.

peter-jerry-ye (commented):
I think this is close, but the escaping boundary and regression test could be tightened.

PR #1871, opened 2026-07-08
feat(cli): support `--deny-warn` in `moon info` and `moon fmt`

co63oc:
- Related issues: None 
- PR kind: 

## Summary

Add the `--deny-warn` (`-d`, "Treat all warnings as errors") flag to `moon info` and `moon fmt`, keeping their behavior consistent with the existing `moon test` / `moon build` commands.

- **`moon info`**: Added a `deny_warn` field and forwarded it into `BuildFlags`, so the RR build engine turns warnings into moonc's deny warning set.
- **`moon fmt`**: Since fmt does not invoke moonc, it now escalates `Warn`-level `user_warnings` from the dependency resolution phase to an error, returning exit code `1` when any warning is present.
- **`moon check --fmt`**: Forward `build_flags.deny_warn` so the fmt call inside check also honors the flag.
- **Docs**: Updated the `moon fmt` options section in `docs/manual` and `docs/manual-zh`.
- **Completion snapshots**: Regenerated the 5 shell completion snapshots (bash/elvish/fish/powershell/zsh).

## Metadata


peter-jerry-ye:
There's no "warnings" so `--deny-warn` is misleading. `moon info` only updates interface files or reports the difference between backends. `moon fmt` only format files without performing any type check. Use `moon fmt --check` instead.

PR #1884, opened 2026-07-10
Respect Why3 environment overrides in `moon prove`

kokic:
This PR fixes moon prove handling of Why3 runtime paths.

  Previously, moon prove always queried:

  - `why3 --print-datadir`
  - `why3 --print-libdir`

  and then temporarily set `WHY3DATA` / `WHY3LIB` process-wide. This can break relocated or binary-distributed Why3
  installations, where why3 --print-*dir may report compile-time paths while the correct runtime paths are provided
  through `WHY3DATA` / `WHY3LIB`.

  This PR changes the behavior to:

  - prefer explicit non-empty `WHY3DATA` / `WHY3LIB`;
  - fall back to why3 --print-*dir only when env overrides are absent;
  - validate the resolved directories early;
  - avoid process-global environment mutation;
  - pass Why3 paths as per-command environment variables only to proof commands.

  It also introduces a dedicated Why3Environment type instead of threading a generic env vector through the build
  plan. This keeps the build-plan API clearer and ensures the environment is only attached to Prove / EmitProof
  commands during lowering.

### DIFF
--- a/crates/moon/src/cli/prove.rs
+++ b/crates/moon/src/cli/prove.rs
@@ -296,11 +296,15 @@ fn configure_why3_environment() -> anyhow::Result<ScopedWhy3Env> {
             "failed to locate `why3` required by `moon prove`; install Why3 1.7.2 preferably via opam, and ensure `why3` is available on PATH"
         )
     })?;
-    let why3_data = query_why3_path(&why3, "--print-datadir", "data")?;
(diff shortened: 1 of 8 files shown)


kokic:
As the final plan is to fully package the why3 related toolchain into the MoonBit toolchain, eliminating the need for users to install why3 themselves, this PR will be closed.

PR #1899, opened 2026-07-15
Remove deprecated --manifest-path option

peter-jerry-ye:
## Summary

- remove the deprecated hidden `--manifest-path` CLI option
- delete the explicit-manifest override field and project-selection branches
- remove the manifest-path handoff to `mooncake`
- update completions, workspace documentation, and affected tests to use `-C`
- retain a regression test that verifies the removed option is rejected

## Why

`-C <project-dir>` is the supported way to select another project. Keeping `--manifest-path` required a second project-selection path, special workspace behavior, and a private serialized field even after the option had been deprecated and hidden.

## Correctness and scope

Project discovery now consistently starts from the working directory after `-C` (or a command-specific selector such as `moon run`). Member-scoped package and publish commands still validate that a workspace member is selected before delegating.

The change removes only the deprecated option and its directly dependent plumbing. Existing project manifest models used by build discovery remain unchanged.

### DIFF
--- a/crates/moon/src/cli/cram.rs
+++ b/crates/moon/src/cli/cram.rs
@@ -234,7 +234,7 @@ fn cram_external_args(raw_args: &[OsString]) -> Option<(Option<PathBuf>, Vec<OsS
             }
         } else if matches!(
             arg.to_str(),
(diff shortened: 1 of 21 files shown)


peter-jerry-ye:
CI failed due to upstream regression

PR #1910, opened 2026-07-17
fix(moonx): keep dependency progress off stdout

peter-jerry-ye:
## Why

Native `moonx` cache misses prepare registry dependencies after source extraction. The default dependency-sync output can print `Downloading` or `Using cached` to stdout before the delegated program starts, corrupting pipelines and command substitution.

## What changed

- Pass explicit dependency-sync output options into native build preparation.
- Keep normal `moonx` dependency acquisition quiet; allow verbose acquisition status on stderr.
- Move registry acquisition status from stdout to stderr.
- Add a local-registry dependency cache-miss regression test proving that `Downloading` is on stderr and delegated stdout remains exact.

## Relationship to #1916

PR #1916 independently deprecates implicit `postadd` hooks and keeps hook output out of automatic and delegated workflows. This PR handles acquisition status only; the branches are intentionally independent so each diff is reviewable against `main`.

## Scope

No archive type, cache/checksum policy, atomic publication, destination replacement, `postadd` policy, or install-flow refactor.

### DIFF
--- a/crates/moon/src/cli/install_binary.rs
+++ b/crates/moon/src/cli/install_binary.rs
@@ -23,7 +23,10 @@ use moonbuild_rupes_recta::{
     intent::UserIntent,
     model::{BuildPlanNode, BuildTarget, PackageId, TargetKind},
 };
-use mooncake::registry::{OnlineRegistry, Registry, path as registry_path};
+use mooncake::{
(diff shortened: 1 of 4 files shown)


peter-jerry-ye:
Superseded by #1918, reopened from the focused two-commit branch so the oversized review history does not carry forward.

PR #1913, opened 2026-07-19
fix(rr): regenerate test driver when patch file changes

hyfdev:
- Related issues: None 
- PR kind: Bugfix 

## Summary

When rerunning `moon test --patch-file` with changed patch content (same patch path, same injected file name), the Rupes Recta build graph rebuilt the injected code but did not regenerate the test driver and `__*_test_info.json` metadata: the `moon generate-test-driver` node did not declare the patch file as an input, while the `moonc build-package` node already did.

This made repeated virtual-test runs silently stale: old test names were reported for new code, and tests added to the injected file never executed while the reported totals still looked green.

Reproducer on current main:

1. Inject one test through `--patch-file ./patch_test.json` and run `moon test -p <pkg> --file <injected>.mbt --patch-file ./patch_test.json` → `Total tests: 1`.
2. Rewrite the patch JSON with two tests under the same injected file name and rerun the same command → still `Total tests: 1` before this fix; `Total tests: 2` after.

The fix tracks `info.patch_file` as an input of the test-driver generation node, mirroring the existing `extend_extra_inputs` handling for compile nodes. The first commit adds a regression test that fails without the fix (the new injected code runs, but the added test never executes); the second commit is the fix plus the five patch dry-run graph snapshots, each gaining exactly the patch file in the driver node's inputs.

## Metadata


peter-jerry-ye:
Thanks

PR #1927, opened 2026-07-23
Share dependency sources for standalone .mbtx runs

peter-jerry-ye:
## Why

Standalone `.mbtx` execution currently unpacks the same registry dependency once per temporary or file-local `.mooncakes` directory. This is especially wasteful for short-lived commands such as `moon run -e` and `moon run -`.

The cache-root and cleaning contract already exists, so this PR adds the first deliberately narrow consumer of the dependency-source cache without changing ordinary project dependency behavior.

## What changed

- Use globally prepared registry sources only for `moon run -e`, `moon run -`, and `moon run <file>.mbtx`.
- Keep project commands, non-`run` single-file commands, and `moon run <file>.mbt` on their existing project/file-local `.mooncakes` flow.
- Resolve the dependency-source mode together with `PackageDirs` before dependency sync, so later phases do not reread process state or recompute directory policy.
- Verify the registry checksum and embedded manifest, extract into staging, make prepared sources read-only, and publish them atomically under a per-module lock.
- Treat `module@version` as the cache identity and reject a changed registry checksum for an already-published version.
- Preserve legacy multi-segment module names as nested cache paths while rejecting unsafe path components.
- Reject source-mutating `postadd` and module prebuild hooks in shared sources.
- Integrate prepared-source removal with `moon clean --dep-cache`.

## Why this is correct

The selected dependency-source mode is an explicit project-resolution fact. Only the three standalone `.mbtx` run entry points select the shared cache; all other callers construct `PackageDirs` with project-local dependency sources.

Readers can observe only a fully verified, immutable source tree. Concurrent first use is serialized, publication is atomic, and a registry checksum change is reported as version immutability corruption rather than creating a hidden second identity. `MOON_DEP_CACHE=off` preserves the file-local fallback.

This branch is based on merged #1928. Its target-owned registry bin-dependency build isolation remains unchanged and is separate from this first shared-source rollout.

## Scope

This is intentionally the first command slice. Migrating additional project or single-file commands can be reviewed independently after the standalone `.mbtx` path has settled.

### DIFF
--- a/crates/moon/src/cli/run.rs
+++ b/crates/moon/src/cli/run.rs
@@ -633,11 +633,18 @@ fn build_single_file_executable_from_arg(
     options: BuildRunExecutableOptions,
     output: &CommandOutput,
 ) -> anyhow::Result<RunExecutable> {
-    let single_file_dirs = cli.source_tgt_dir.single_file_package_dirs(
-        cmd.package_or_mbt_file
-            .as_deref()
-            .expect("single-file run from arg requires a positional input path"),
-    )?;
+    let input = cmd
+        .package_or_mbt_file
+        .as_deref()
+        .expect("single-file run from arg requires a positional input path");
+    let single_file_dirs = if Path::new(input)
+        .extension()
+        .is_some_and(|extension| extension == "mbtx")
+    {
+        cli.source_tgt_dir.cached_single_file_package_dirs(input)?
+    } else {
+        cli.source_tgt_dir.single_file_package_dirs(input)?
+    };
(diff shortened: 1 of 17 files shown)


peter-jerry-ye:
Closing this version so the source-cache seam can be refactored and the work can be re-cut into vertically scoped PRs. A replacement will keep prepared dependency sources independent from the follow-up dependency build cache.

peter-jerry-ye:
Replacement split is now available as #1935 (prepared dependency sources) and #1936 (stacked dependency build cache).

PR #1931, opened 2026-07-24
Normalize UserLog levels and build result ownership

peter-jerry-ye:
## Why

Build output had two coupled problems: the default `UserLog` level hid normal lifecycle messages, and Rupes Recta execution mixed compilation with durable command-result presentation. That made build results inconsistent, treated explicit skipped inputs like internal detail, and allowed build summaries to compete with compiler diagnostics on stderr.

## What

- Normalize shared user-log filtering to `Error` for `--quiet`, `Info` by default, and `Debug` for `--verbose`; `Info` and `Debug` both remain bare text.
- Keep normal lifecycle messages such as installs, downloads, clones, and registry updates at `Info`.
- Reclassify cache hits, concrete command echoes, and optional backend-selection detail as `Debug`.
- Report explicitly requested paths that are skipped as `Warn`, while keeping automatic workspace filtering silent by default.
- Make RR build execution return statistics without owning the durable build result. Full command callers decide whether and how to report it.
- Write human `Finished ...` and `Failed with ...` build results to stdout through `CommandOutput`.
- Apply `--quiet` only to successful status output; failed build results remain visible.
- Preserve the semantic build exit status when a pipe consumer closes stdout early; other stdout write failures still propagate.
- Keep failure context in `UserLog::error` and compiler diagnostics in their existing independent renderer and channel policy.
- Omit intermediate build results for commands that own a later primary result, including `run`, `prove`, tests, benchmarks, and internal moonx builds.
- Keep JSON diagnostic stdout free of human summary text.
- Preserve the current successful exit behavior when all explicit inputs are skipped.

## Why this is correct

Each output now has an explicit owner: `CommandOutput` owns command-result and command-status stdout, `UserLog` owns filtered Moon-authored stderr messages, and the diagnostic renderer continues to own compiler diagnostics. Success and failure summaries follow the same channel rule, while quiet filtering remains a presentation policy rather than a log-level side effect. Build execution still emits diagnostics and progress as before, but callers decide whether a durable build result belongs to the command. A regression test closes the stdout pipe before both successful and failed summaries are written, proving that output delivery cannot replace exit codes 0 and 1 with a BrokenPipe error.

## Scope

This PR is limited to build-result ownership and persistent user-log filtering. New command-level `--json` and `--jsonl` formats remain isolated in the separate structured-output work, and changing the all-explicit-inputs-skipped exit behavior remains follow-up work.

### DIFF
--- a/crates/moon/src/cli/build.rs
+++ b/crates/moon/src/cli/build.rs
@@ -204,7 +204,6 @@ fn run_build_rr(
             if result.successful() {
                 crate::resources::reconcile_resource_mappings(&build_meta)?;
             }
-            result.print_info(cli.quiet, "building")?;
             ok &= result.successful();
         }
         ok
--- a/crates/moon/src/cli/build.rs
+++ b/crates/moon/src/cli/build.rs
@@ -204,6 +204,7 @@ fn run_build_rr(
             if result.successful() {
                 crate::resources::reconcile_resource_mappings(&build_meta)?;
             }
+            rr_build::report_build_result(&result, rr_build::BuildOperation::Build, &cfg, output)?;
             ok &= result.successful();
         }
         ok
--- a/crates/moon/src/cli/bundle.rs
+++ b/crates/moon/src/cli/bundle.rs
@@ -135,7 +135,6 @@ pub(crate) fn run_bundle_internal_rr(
             target_dir,
             user_log,
         )?;
-        result.print_info(cli.quiet, "bundling")?;
         Ok(result.return_code_for_success())
     }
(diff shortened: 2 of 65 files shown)


peter-jerry-ye:
Closing this in favor of a smaller, incremental split: first characterize and document the existing moon check output contract, then inject the compiler diagnostic writer without changing the current stdout/stderr behavior. This avoids mixing UserLog policy, build-result ownership, stream changes, and widespread snapshot updates in one PR. The replacement PRs will be opened separately.

PR #1935, opened 2026-07-24
Share prepared dependency sources for standalone scripts

peter-jerry-ye:
## Why

Standalone `.mbtx` execution currently unpacks the same registry dependency once per temporary or file-local `.mooncakes` directory. Short-lived commands such as `moon run -e` and `moon run -` therefore repeat source preparation before they can reuse any dependency build artifacts.

This PR establishes one stable prepared-source seam for the standalone path. It also exposes the verified registry checksum needed by a follow-up compiler-artifact cache without coupling that cache to the on-disk source layout.

## What changed

- Use globally prepared registry sources for `moon run -e`, `moon run -`, and `moon run <file>.mbtx`.
- Keep ordinary projects, other single-file commands, and `moon run <file>.mbt` on their existing project-local `.mooncakes` flow.
- Move the shared source-cache lifecycle into a dedicated module that returns `{ source_dir, verified_checksum }`.
- Verify the registry archive and embedded manifest, reject source-mutating `postadd`, publish through staging under a per-module lock, and make the complete entry read-only.
- Recursively reject writable nodes, symlinks, and special files on every cache hit.
- Preserve frozen resolution as a read-only operation and integrate the cache with explicit dependency-cache cleaning.

## Why this is correct

A registry `module@version` remains one immutable source identity. The recorded checksum must match the current registry checksum; a mismatch is reported as registry corruption rather than creating a hidden second variant. Readers can observe only a complete, recursively validated source tree, and concurrent first use is serialized before atomic publication.

`MOON_DEP_CACHE=off` preserves the existing file-local fallback. Commands outside the three standalone `.mbtx` forms do not opt into the new source location.

## Scope

This PR only prepares immutable dependency sources. It does not publish compiler outputs, share an n2 database, move project `packages.json`, or change mutable `_build` ownership. Dependency build reuse is split into a follow-up PR that consumes the checksum interface introduced here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1727,6 +1727,7 @@ dependencies = [
  "tempfile",
  "test-log",
  "thiserror 1.0.63",
+ "tracing",
  "walkdir",
  "zip",
 ]
@@ -1793,6 +1794,7 @@ dependencies = [
  "slotmap",
  "tempfile",
  "thiserror 1.0.63",
+ "tracing",
  "vergen",
  "which 6.0.2",
 ]
--- a/crates/moon/src/cli/run.rs
+++ b/crates/moon/src/cli/run.rs
@@ -633,11 +633,18 @@ fn build_single_file_executable_from_arg(
     options: BuildRunExecutableOptions,
     output: &CommandOutput,
 ) -> anyhow::Result<RunExecutable> {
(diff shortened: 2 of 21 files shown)


peter-jerry-ye:
Closing as superseded. The dependency-cache work is being rebuilt from the merged lowering and tool-input boundaries rather than continued on this draft.

PR #1936, opened 2026-07-24
Cache registry dependencies for standalone scripts

peter-jerry-ye:
## Why

Standalone `.mbtx` execution currently repeats two kinds of stable work: preparing the same registry sources in a temporary `.mooncakes` tree and compiling the same resolved dependency graph into invocation-local outputs. Shipping only one half has limited value: source reuse alone still recompiles dependencies, while artifact reuse needs a stable verified source identity.

This PR delivers the complete narrow MVP for `moon run -e`, `moon run -`, and `moon run <file>.mbtx`: globally share immutable registry sources and validated dependency build outputs while keeping every mutable project workspace private.

### Prepared dependency sources

- Add separate dependency-source and build-artifact cache roots with explicit `off` semantics and safe `moon clean --dep-cache` / `--build-cache` ownership checks.
- Prepare registry dependencies once under `module@version`, verifying the archive checksum and embedded manifest before atomic publication.
- Return the prepared source directory together with its verified checksum through a dedicated mooncake module.
- Recursively reject writable nodes, symlinks, and special files on cache hits; reject source-mutating `postadd` hooks.
- Preserve frozen resolution as a read-only operation and keep ordinary projects, other single-file commands, and `.mbt` execution on their existing local dependency flow.

### Dependency build outputs

- Collect registry dependency `BuildCore`, C-stub compile, and C-stub archive actions from normal Rupes Recta lowering; include the native runtime action where applicable.
- Identify the complete resolved dependency subgraph with SHA-256 over action kinds, structural command argv, target/options, explicit command environment, exact upstream resolution, logical product edges, each prepared archive checksum, and content digests for external inputs.
- Hash the actual contents of `moonc`, the selected C compiler, and the selected archiver. Persist metadata-keyed SHA-256 memos for large inputs so unchanged tools are not reread for every invocation while their content digest remains in the graph identity.
- On a miss, let one n2 graph build the dependency outputs with normal parallelism, then publish the complete reusable result under a graph lock.
- On a hit, validate every recorded output before restoring it into the invocation-private target directory and detaching only the restored producer nodes from the local graph.
- Cache Native and LLVM dependency C-stub objects and archives plus the runtime library as part of the same complete result. Final script linking remains local.

## Why this is correct

A registry `module@version` remains one immutable prepared-source identity. Its recorded checksum must match the current registry checksum; a mismatch is reported as corruption rather than creating a hidden second variant. Readers observe only complete, recursively validated source trees.

Build artifacts have a stronger identity than `module@version`: the prepared checksums cover source/configuration bytes, exact resolution facts cover selected upstream versions, and the structured actions cover target, arguments, environment, tools, and dependency-product edges. Tool contents, exact arguments, and explicit command environment make compiler, CC, and archiver changes produce a different graph identity.

Source and artifact publication are serialized through separate locks and staging directories. Missing, malformed, symlinked, size-mismatched, or digest-mismatched outputs are treated as misses and rebuilt once under the graph lock.

## Scope

This remains deliberately limited to the three standalone `.mbtx` run forms. The artifact unit is one complete registry dependency subgraph, not a per-package or per-action cache. `packages.json`, the n2 database, mutable `_build` work, the script package, and final linking remain invocation-local. Partial-overlap reuse, pruning, and migration of other commands are follow-up work.


peter-jerry-ye:
Closing as superseded. The dependency-cache work is being rebuilt from the merged lowering and tool-input boundaries rather than continued on this draft.

PR #1948, opened 2026-07-27
Cache single-file dependency action outputs locally

peter-jerry-ye:
## Why

Single-file dependency preparation should reuse outputs without making n2's
persistent database the owner of dependency freshness. The cache also needs to
use the lowering model directly instead of reconstructing build semantics from
an already-rendered n2 graph.

## What changed

- Rupes Recta retains dependency `LoweredAction`s while continuing to lower the
  synthesized script into its normal n2 graph. Ordinary project and workspace
  builds continue to stream actions directly into n2.
- Moon derives canonical action digests directly from structured argv, cwd,
  the complete effective environment, response-file path and contents, sorted
  external inputs, recursive producer digests with logical products and
  concrete paths, and sorted output paths.
- `BuildActionId` is used only to find producers within the current lowering.
  Numeric action IDs are not persisted, and the cache neither reads nor copies
  an n2 graph.
- Concrete compiler and archiver executables, MoonLex and MoonYacc programs,
  source files, package-local C headers, and the standard-library interface
  tree are explicit lowered external inputs. Standard-library symlinks are
  followed with cycle protection.
- C-stub caching uses a conservative package-file-set boundary, like Go's cgo
  cache: all recognized headers below the package root participate, while
  headers and libraries reached only through external search paths remain
  external toolchain state.
- Cache hits validate a content-addressed manifest and every output before
  materializing the complete output set. Selected misses enter n2 only through
  the narrow `LoweredAction` adapter and use per-invocation executor state.
- Object publication tolerates duplicate and concurrent writers. A loser
  validates the winner, while malformed entries, damaged manifests, missing
  files, and partial objects become safe misses.

## Why this is correct

The identity follows producer relationships transitively without depending on
graph insertion order or hash collection iteration order. Publishing a complete
output object before its action entry prevents false hits, and external inputs
are revalidated before successful execution results are published.

Retained dependency actions snapshot their inherited environment and apply
action-specific overrides before both identity and execution. This conservative
contract covers arbitrary compiler wrappers now; a later controlled minimal
environment can narrow both sides without changing the action-to-output model.

The store's publish and restore operations tolerate local races. This does not
claim that independent Moon processes may concurrently mutate the same
`_build`; that broader behavior still relies on the existing target-directory
lock.

## Commit structure

1. retain dependency actions before n2;
2. add the lowered-action identity and local output store;
3. cover end-to-end behavior and store corruption/concurrency;
4. complete file-input coverage;
5. capture the effective dependency action environment;
6. document the cache input boundary.

The change remains limited to single-file `.mbt` and `.mbtx` dependency
preparation. It does not enable a workspace-wide cache or perform a broad
standalone-to-single-file naming cleanup.


peter-jerry-ye:
Closing as superseded by the merged lowering/toolchain prerequisites. The cache will be rebuilt as a smaller change from current main, using LoweredAction as its source of truth.

PR #2005, opened 2026-08-07
Revert "Avoid unstable packages.json updates from focused checks"

bzy-debug:
Reverts moonbitlang/moon#2004


bzy-debug:
`moon ide doc` relies on pre-bunlded core to genreate package.json.

After we make `moon ide doc` run moon check on core, we can re-revert this

PR #2011, opened 2026-08-08
fix: skip missing toolchain companions in dry-run path normalization

jayphelps:
Essentially I need to use moonbit on an Intel Mac, so I have to use the wasm build of `moon` but can't because `moon test` expects all the other non-wasm available commands to be available.

## Summary
`moon test` logs each launch command through `PathNormalizer`, which previously forced `all_moon_bins()` and therefore resolved every companion binary (`mooncake`, `moondoc`, `moon_cove_report`, …). On incomplete installs—especially the **Wasm toolchain**, which does not ship those native companions—`moon test` panicked while merely formatting the command line, before any test ran.

This change resolves companions **opportunistically** for the dry-run / log replace table. Actual tool use still panics via the existing `moon_executable` / `moon_payload` paths when a required binary is missing.

## Test plan
- [x] `cargo test -p moonutil binaries` (including new `try_moon_executable` / `all_moon_bins_skips_uninstalled_companions`)
- [x] `cargo test -p moonbuild normalizes_`
- [x] Against an incomplete Wasm toolchain install (no `~/.moon/bin/mooncake`):
  - stock `moon test --frozen` in a real project → panics on `mooncake`
  - this branch’s `moon` → builds and **launches** all tests (no resolve panic)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/moonbuild/src/dry_run.rs
+++ b/crates/moonbuild/src/dry_run.rs
@@ -113,6 +113,9 @@ fn normalized_env_lines(env: &[(String, String)], replacer: &PathNormalizer) -> 
 // FIXME: `PathNormalizer` is production-facing dry-run output formatting, not
 // moonbuild debug support. Move it to a non-debug utility module after
 // `moonbuild-debug` is no longer needed on production dependency paths.
+//
+// `all_moon_bins()` intentionally skips missing companions so formatting a
+// `moon test` command line does not require `mooncake` / `moondoc` / etc.
 pub struct PathNormalizer {
     canonical: Option<PathBuf>,
     replace_table: Vec<(String, String)>,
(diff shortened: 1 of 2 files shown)


peter-jerry-ye:
Thank you for your contribution. I can see your frustration and I admit that the `PathNormalizer` is poorly implemented.

However, I'm also not very fond of the change in this PR. It feels we are just making patches here and there. I think what I would like to do is to make the `PathNormalizer` best effort only. 

Its purpose was to make a stable, readable, privacy safe snapshot. It tries to hide away user's path (which usually include username and other stuff), stablizes the snapshot, which is used a lot in this repo, and normalize the path separator.

Thus I think I would prefer to make `PathNormalizer` detect only the toolchain path + the overriden environment variables so that it gets decoupled with the actual bin resolution.

I think I'll draft an alternative PR later, but should certainly solve your issue.

peter-jerry-ye:
Replaced with #2013. I shall remove `all_moon_bins` in another follow up PR

PR #2022, opened 2026-08-11
refactor(build): share target-directory n2 state

peter-jerry-ye:
## Why

Rupes Recta currently selects a separate n2 database for each backend, profile,
run mode, and standalone build phase. That prevents separately planned graphs
from reusing execution state for actions with the same physical outputs, such
as package prebuilds, and would require a second cache layout solely for
multi-backend graph composition.

## What changed

- use one `.moon_db` per target directory for Rupes Recta executions
- remove backend-, profile-, run-mode-, and standalone-phase database selection
- serialize `moon info` and `moon fmt` through the existing target-directory lock
- cover reuse across backend, profile, and run-mode boundaries
- keep standalone dependency and script graphs sequential while sharing their
  persistent n2 state

## Why this is correct

n2 associates records with concrete output paths and validates the current
build hash before treating an action as up to date. Backend, profile, and
run-mode differences already affect those paths or hashes, so they do not need
separate database files. Moon holds the target-directory lock while executing
these graphs because n2 does not lock the database internally.

Existing per-configuration databases are not migrated, so the first execution
after this change may rebuild once.

## Scope

This PR only unifies persistent executor state. Backend Build Plans and n2
graphs remain separate, and standalone dependency and script graphs still run
sequentially. The n2 dependency remains unchanged and its database remains
append-only; automatic compaction and multi-graph-safe retention are follow-up
work.

### DIFF
--- a/crates/moon/src/cli/fmt.rs
+++ b/crates/moon/src/cli/fmt.rs
@@ -20,7 +20,7 @@ use std::path::PathBuf;
 
 use anyhow::Context;
 use moonbuild_rupes_recta::fmt::FmtConfig;
-use moonutil::{command_output::CommandOutput, project::PackageDirs};
+use moonutil::{command_output::CommandOutput, locks::FileLock, project::PackageDirs};
 
 use crate::filter::{filter_pkg_by_dir_for_fmt, select_packages};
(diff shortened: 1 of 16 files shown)


peter-jerry-ye:
We might postpone this because n2 doesn't currently have auto compact

PR #2025, opened 2026-08-11
fix(dry-run): normalize complete path tokens

peter-jerry-ye:
## Why

Dry-run path masking handled configured binary overrides only as whole arguments and recognized directory roots mainly by their trailing boundary. Nested shell commands could therefore retain a resolved `moon` path, while textual prefixes could be mistaken for directories, making output depend on host-specific absolute paths.

## What changed

- Replace configured binaries and the running `moon` path only when they form complete shell tokens, including inside nested command payloads.
- Require both leading and trailing boundaries when masking directory roots, with support for path lists and path-bearing compiler options.
- Keep environment-variable aliases as symbolic literals in the displayed command.
- Document the implemented dry-run masking rules.

## Why this is correct

The matching rules distinguish exact file tokens from directory prefixes and cover POSIX and Windows command shapes, compact compiler flags, linker path options, false textual prefixes, and the `:embed` prebuild expansion.

The change is limited to dry-run display normalization; command execution and build planning are unchanged.

### DIFF
--- a/crates/moon/tests/test_cases/prebuild/mod.rs
+++ b/crates/moon/tests/test_cases/prebuild/mod.rs
@@ -144,6 +144,38 @@ fn test_pre_build_dry_run_uses_cwd_relative_placeholders() {
             .any(|line| line.contains(" -i ./src/lib/a.txt -o ./src/lib/a.mbt")),
         "dry-run should show cwd-relative placeholders, got:\n{stdout}"
     );
+    assert!(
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
I think a lot of issues are brought by the fact that we reconstruct dry-run commands from the n2 graph, which brought a bunch of complexity. So I think we will try to refactor the dry-run instead.

PR #2034, opened 2026-08-12
chore: update warning index snapshot

peter-jerry-ye:
## Summary

The compiler warning index now includes warning 90, `unused_lexcase`, causing the checked-in warning snapshot to fall out of sync with `moonc check -warn-help`.

This PR adds the same warning entry to Moon's warning table and refreshes the corresponding snapshot. The change is limited to the synchronized table and snapshot.


peter-jerry-ye:
A flaky issue is found but is not blocking.

PR #2061, opened 2026-08-17
refactor(build): retain package interface inputs in build plan

peter-jerry-ye:
## Why

Package interface dependencies were represented twice: Build Plan selected lifecycle-specific artifacts and providers, while lowering traversed Resolve Output again to reconstruct `.mi` paths, aliases, and virtual-package validation. Those independent paths could drift; virtual implementations already exposed the problem by inserting the same contract file into the execution graph twice.

## What changed

- retain an action-specific Package Compilation Specification after Build Target Projection, including interface Artifact Keys, import aliases, and regular/default-virtual/implementation behavior
- let the Artifact Registry resolve each requirement to either a planned provider or a declared external source such as the injected standard library
- derive interface requirements from that specification and make compiler lowering consume only the specification plus realized artifact paths
- preserve the recursive standard-library observation and the exact external virtual-contract observation without duplicating locally produced contract edges
- update the virtual-package graph expectations, glossary, and architecture references

## Why this is correct

Package compilation behavior now has one semantic source in Build Plan. Check, build, proof, and virtual-contract interfaces remain distinct Artifact Keys; provider selection stays in the Artifact Registry; lowering cannot independently choose another interface path. The compiler command is unchanged, while the duplicated virtual-contract execution input is removed.

## Scope

This adds no execution layer or persistent action identity. Proof-specific `--dep-proof` metadata and transitive WhyML load paths still read Resolve Output and are marked as follow-up work, as is generated-source provenance. The change is limited to package interface planning and lowering.


peter-jerry-ye:
Superseded by #2065, which replaces the produced-input classification with the agreed provider-topology model and removes the obsolete PrebuildOutput representation.

PR #2085, opened 2026-08-21
feat(native/windows): auto-select C toolchain (MSVC, Clang, MinGW)

Lfan-ke:
On Windows, native backend currently only supports MSVC (cl.exe / clang-cl.exe). Users with MinGW or a standalone Clang have to set MOON_CC manually. This changes the detection order to MSVC > MinGW > system cc so all three work out of the box.

Three changes:

`detected_default_native_toolchain` on Windows - after MSVC probe fails, explicitly try `x86_64-w64-mingw32-gcc` and `i686-w64-mingw32-gcc` before falling through to the generic system-cc probe. Canonical MinGW names get priority over a bare `gcc` that may point to something else.

`NativeTarget::from_env_for_host` - when `MOONBIT_NEW_NATIVE=1` on Windows but MSVC is absent, return `None` instead of `X86_64PcWindowsMsvc`. The build falls back to generated-C mode, which then picks up MinGW/GCC through the detection above. The direct-object path cannot support MinGW because it emits MSVC-ABI objects; the generated-C path works with any C compiler.

Error and warning messages in `builders.rs` updated to mention that MinGW/GCC is supported via generated-C mode and how to reach it.

Existing Linux/macOS behaviour is unaffected (`#[cfg(windows)]` guards). The `from_host_with_new_native_env` tests cover the original behaviour and still pass without modification. Windows + MinGW integration testing requires a real MSYS2 environment not available in CI.

### DIFF
--- a/crates/moonbuild-rupes-recta/src/build_plan/builders.rs
+++ b/crates/moonbuild-rupes-recta/src/build_plan/builders.rs
@@ -87,7 +87,8 @@ impl<'a> BuildPlanConstructor<'a> {
     fn new_native_linker_context(&self, err: anyhow::Error) -> anyhow::Error {
         if self.build_env.direct_native_target() == Some(NativeTarget::X86_64PcWindowsMsvc) {
             err.context(
-                "Windows MSVC direct object native target requires an MSVC compiler/linker driver such as cl.exe or clang-cl.exe",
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
We do not support MinGW at the moment, and the default binary output on windows support MSVC only.

PR #2091, opened 2026-08-24
feat(tree): add machine-readable and package-level dependency views

kokic:
`moon tree` gains machine-readable and package-level dependency views,
following the Command Result contract established by `moon check --json`
(one compact JSON document on stdout, captured user logs embedded in the
document, stderr stays empty):

- `moon tree --json`: the resolved module dependency graph — `modules`
  (name, version, source kind + path/url, workspace_member), `edges`
  (from/to indices, dependency name, regular/binary), selected `root`,
  plus a versioned `{status, error, logs}` envelope.
- `moon tree --package`: the package-level dependency graph as a text
  tree, rooted at the selected module's source packages.
- `moon tree --package --json`: the same package-level graph as JSON —
  packages deduplicated by identity (`module`, `version`, `source`,
  `rel`) as nodes, and import edges with their short aliases and the
  aggregated target kinds they originate from; standard-library
  packages are excluded.

Tree presentation moved out of mooncake (CO-6): the resolved graph is
returned as data and both views are rendered through `CommandOutput`.


peter-jerry-ye:
Thanks for moving the tree presentation to the CLI seam. I reviewed this mainly for code structure rather than the inherited tree expansion behavior; I do not consider the existing shared-subtree repetition attributable to this PR.

I think the following should be addressed:

1. **The JSON lifecycle now has repeated command switches.** JSON selection is handled in `cli/invocation.rs:172-178`, then `cli/runtime.rs` separately switches over `Check` and `Tree` to run the command, construct bootstrap failures, retrieve exit codes, and select a renderer (`:168-245`). Adding another machine-readable command requires coordinated edits in five places, with missed cases becoming `unreachable!` panics. This is the mode-switched control flow that `AGENTS.md` asks us to replace with injected varying behavior. Please select command-owned run/bootstrap-error/finish hooks once and keep runtime responsible only for capture, tracing, and finalization.

2. **The module and package text renderers duplicate the traversal algorithm.** `render_package_tree_edges` (`tree.rs:536-587`) and `render_tree_edges` (`:651-701`) independently implement sorting, branch glyphs, indentation, cycle handling, stack management, and recursion. The inherited expansion policy is fine to preserve here, but the new copy creates a synchronization point for every future rendering change. Please extract one generic traversal scaffold and inject graph-specific sorted-child/label behavior.

3. **`package_mode: bool` encodes the JSON failure schema as a mode flag.** `TreeJsonOutcomeKind::Failure` (`tree.rs:141-165`) can represent a package failure with the module schema or vice versa, and `write_tree_json` has to reconstruct the association manually. A structural shape such as `Module(Result<...>)` / `Package(Result<...>)` would make invalid combinations impossible. Relatedly, `tree.rs` imports `runtime::INTERNAL_ERROR_EXIT_CODE`; runtime policy should remain private and the outcome should carry an explicit exit code.

There is also one observable regression: plain `moon tree` now adds `[workspace member]` for a single-member workspace. Before this PR, the human marker was emitted only with multiple input roots. The CO-6 migration says existing behavior should remain stable unless an intentional change is named and tested, while accurate membership is specifically required for JSON. Please preserve the old human condition independently from the JSON `workspace_member` field.

Smaller follow-ups: move the CLI-only `TreeSubcommand` flags out of mooncake; assert JSON stdout and stderr from the same process rather than two executions; and add a package-edge test covering target-kind aggregation and distinct aliases.

PR #2107, opened 2026-08-26
feat(check): publish package metadata backend index

peter-jerry-ye:
## Why

Package metadata consumers can select one active scoped document through
`_build/packages.json`, but they cannot discover every backend published by a
multi-backend check. They need a small, authoritative backend inventory without
another project discovery or resolution pass.

## Why this is correct

- `_build/index.json` is derived from the exact check plans that publish the
  scoped metadata documents, so explicit targets, configured preferences, and
  supported-target filtering are already reflected.
- Shared check finalization publishes all scoped documents, then the selector,
  and finally one accumulated index. This covers project and standalone checks,
  including multi-target standalone invocations.
- Backend entries are deduplicated and deterministic. Focused checks and other
  commands that do not own this metadata leave the existing index untouched.

## Scope

This PR adds the index path, publication lifecycle, behavioral coverage, and
reference documentation. It does not change the scoped `packages.json` schema
or the universal selector's two-field shape.

### DIFF
--- a/crates/moon/src/cli/check.rs
+++ b/crates/moon/src/cli/check.rs
@@ -419,7 +419,7 @@ fn run_check_impl(
                 .collect()
         };
         let mut ret_value = 0;
-        let mut target_backends = Vec::with_capacity(targets.len());
+        let mut build_metas = Vec::with_capacity(targets.len());
         for t in targets {
             let result = run_check_for_single_file_rr(
                 cli,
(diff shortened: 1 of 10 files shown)


peter-jerry-ye:
Splitting this work as discussed. Standalone Check plan composition has moved to #2108. Closing the backend-index PR until that refactor lands; its feature branch is preserved for rebase and reopening afterward.

Issue #2118, opened 2026-08-28
moon doc plans moonc checks differently from moon check


Yoorkin:
## Description

Before invoking `moondoc`, `moon doc` runs `moonc check` to prepare package metadata and interface files. The packages checked in this phase are not selected consistently with `moon check` for the same target backend.

`moon check` selects local packages that support the requested backend and expands only the dependencies needed by those packages. In contrast, `BuildDocs` currently requests `CheckMi` for every discovered non-stdlib package through `all_packages(true)`. This includes unrelated dependency packages and packages whose `supported_targets` excludes the selected backend.

## Reproduction

Using `moonbit-community/rabbita` with `moonbitlang/async@0.21.0`:

```console
$ moon doc --target wasm --dry-run | rg 'async/src/raw_fd/raw_fd.mbt'
moonc check .../async/src/raw_fd/raw_fd.mbt ... -target wasm ...

$ moon check --target wasm --dry-run | rg 'async/src/raw_fd/raw_fd.mbt'
# no output
```

The package explicitly supports only the native backend:

```moonbit
supported_targets = "-all+native"
```

Executing the commands produces the same difference:

- `moon check --target wasm` succeeds.
- `moon doc --target wasm` checks `moonbitlang/async/raw_fd` and fails because `fd_is_nonblocking` is not available for Wasm.
- The JS target also checks unrelated packages such as `hackwaly/moonback` and `moonbitlang/async/raw_fd`, producing undefined-type/value errors, while `moon check --target js` succeeds.

The dry-run package-set comparison showed 27 packages checked only by `moon doc` for Wasm and 36 packages checked only by `moon doc` for JS. Raw action totals are not directly comparable because `moon check` also creates test and doctest check actions.

## Relevant implementation

- `crates/moonbuild-rupes-recta/src/build_plan/builders.rs`: `build_build_docs` requests source `CheckMi` for `pkg_dirs.all_packages(true)`.
- `crates/moon/src/cli/check.rs`: normal `moon check` intents start from local packages filtered by `package_supports_backend`.

## Expected behavior

The compiler-check phase used to prepare documentation should respect the selected backend and should not check packages excluded by `supported_targets`. Ideally, its package selection should share the applicable behavior with `moon check`; any intentional documentation-specific differences should be explicit.

PR #2119, opened 2026-08-28
feat(doc): add target selection

Yoorkin:
## Summary

- add `--target <TARGET>` to `moon doc`
- pass the selected backend into the existing documentation check pipeline
- preserve the existing target resolution when `--target` is omitted
- cover explicit target selection in the dry-run and Windows long-path tests
- update the generated English and Chinese command references

This PR intentionally adds only `--target`; it does not expose the other build flags used by commands such as `moon check`.

## Testing

- `cargo test -p moon --test mod test_moon_doc_dry_run_with_target -- --nocapture`
- `cargo clippy -p moon --all-targets -- -D warnings`

## Related issue

The existing documentation build plan checks a different package set from `moon check` before the final `moondoc` invocation. That behavior is separate from adding target selection and is tracked in #2118.

Related to #2118.

### DIFF
--- a/crates/moon/src/cli/doc.rs
+++ b/crates/moon/src/cli/doc.rs
@@ -25,6 +25,7 @@ use moonutil::command_output::CommandOutput;
 use moonutil::constants::MOON_MOD_JSON;
 use moonutil::project::PackageDirs;
 use moonutil::resolution::ModuleId;
+use moonutil::target::TargetBackend;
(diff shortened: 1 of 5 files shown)


peter-jerry-ye:
This is due to the missing of the backend selection based on preferred target. Related to #2118 where the moon doc is not in sync.

PR #2132, opened 2026-08-31
enable `/std:c11` for MSVC

Guest0x0:
Currently, no `/std` flag is passed when using MSVC to compile things (runtime, stub, `moonc` generated C code). This correspond to C89 + MSVC extension according to Microsoft's doc. This PR pass `/std:c11` when compiling stuff using MSVC to enable several modern C features, such as `stdatomic.h`.


peter-jerry-ye:
what about clang / gcc?

PR #2220, opened 2026-09-16
feat: support conditional imports in moon.pkg

Yu-zh:
- Related issues: None
- PR kind: Feature

## Summary

Adds backend-conditional imports to `moon.pkg` for regular, blackbox-test, and whitebox-test imports. Conditions support `target = "..."`, `all(...)`, `any(...)`, `not(...)`, `true`, and `false`, with clearer diagnostics for misplaced `#cfg` attributes and unsupported conditions.

Package discovery happens before backend selection. Dependency resolution produces complete `PackageRelations` for each requested backend, which is shared by planning, lowering, metadata, package-graph export, and virtual-package handling. For example, `moon check --target native` ignores broken JS-only imports, while `moon check --target js,native` fails if either requested backend cannot resolve. Run planner tests use the production package and backend selection path.

Builds on the merged parser, normalization, direct DSL-to-MoonPkg conversion, and discovery/resolution split. Uses the merged `DiscoveredProject` and `ResolvedProject` naming, with coverage configuration retained by discovery. Repeated import blocks combine in declaration order. Repeated packages within one import kind remain valid and warn only when their conditions overlap. Legacy import options override the corresponding blocks before warnings; aliases and import-all settings retain their existing behavior. Legacy `moon.pkg.json` imports remain unconditional.

Includes package-model, dependency-graph, planner, and CLI coverage, plus English and Chinese documentation.

## Validation

Rebased onto `main` at `e46d2ed38` (including merged #2241, #2242, and #2243). The final update from `86c63700e` changes only `moonrun`; `git range-diff` confirms all five PR patches are unchanged.

Validation before the final runtime-only base update:

- `cargo test --locked -p moonutil -p moonbuild-rupes-recta` — 407 tests passed; one existing doctest ignored.
- `cargo test --locked -p moon --bin moon --test mod -- tests::planner filter::tests watch::prebuild_output::tests conditional_imports major_version_modules_coexist_in_projects_and_scripts test_wbtest_coverage` — 59 unit tests and 9 CLI tests passed, including Wasm execution for versioned imports in projects and scripts, plus coverage execution.
- `cargo clippy --workspace --exclude moonrun --locked --all-targets --all-features -- -D warnings`
- `git diff --check`
- Generated-C test planning passed with `MOONBIT_NEW_NATIVE=0 moon test app --target native --dry-run` on the conditional-import fixture.

After the final base update, reran `cargo test --locked -p moon --test mod -- major_version_modules_coexist_in_projects_and_scripts test_wbtest_coverage`: both integration tests passed, covering ordinary project/script Wasm execution and coverage. Formatting and diff checks also passed again.

Native/C checks used dry-runs only; the full workspace test suite was not rerun.

Actual Wasm execution of the conditional-import fixture was attempted, but the installed compiler (`moonc v0.10.14+7d59c7ec9`, 2026-09-18) rejects `#cfg` in `moon.pkg`. Compiling and formatting conditional manifests still requires compiler/formatter support for that syntax. Ordinary project and script Wasm execution passed as described above.

## Metadata

- [ ] Compatible with Windows/Linux/macOS (focused macOS validation; cross-platform validation is left to CI)


crates/moonutil/src/package.rs
```diff
@@ -807,6 +818,18 @@ pub(crate) fn convert_pkg_dsl_to_package_with_supported_targets_decl(
...
+                Value::Array(values)
+                    if matches!(key, "import" | "wbtest-import" | "test-import") =>
+                {
+                    import_keys.insert(key);
+                    values.extend(
+                        value
+                            .as_array()
+                            .expect("import block is an array")
+                            .iter()
+                            .cloned(),
```
peter-jerry-ye: [P2] Reject duplicate active imports before graph insertion

Concatenating repeated blocks currently loses the opportunity to validate overlapping declarations. For example:

```moonbit
#cfg(any(target = "native", target = "js"))
import { "example/lib/common" @first }

#cfg(target = "native")
import { "example/lib/common" @second }
```

On Native, both imports are active, but the dependency graph retains only `@second`; reversing the blocks retains only `@first`. The existing graph-edge replacement behavior therefore makes the newly accepted block combinations depend on declaration order.

Multiple blocks are allowed, but each imported package must appear at most once per backend within each import kind, even if its aliases differ. Check for intersecting backend sets before inserting graph edges. Please cover overlapping conditions, unconditional duplicates, and identical/different aliases; disjoint conditions and different import kinds should remain valid. The new `conditional_imports_select_aliases_and_propagate_backend_support` test currently expects this alias replacement and should instead reject the duplicate.

crates/moonutil/src/package.rs
```diff
@@ -864,9 +888,29 @@ pub(crate) fn convert_pkg_dsl_to_package_with_supported_targets_decl(
...
+        .map(|key| {
+            serde_json_lenient::from_value::<Vec<Import>>(map[key].clone())
+                .map(|imports| (key, imports))
+        })
+        .collect::<Result<std::collections::HashMap<_, _>, _>>()?;
     let json = Value::Object(map);
     let pkg_json: MoonPkgJSON = serde_json_lenient::from_value(json)?;
-    convert_pkg_json_to_package_with_supported_targets_decl(pkg_json, emit_warnings, user_log)
+    let (mut pkg, supported_targets_decl) =
+        convert_pkg_json_to_package_with_supported_targets_decl(pkg_json, emit_warnings, user_log)?;
```
peter-jerry-ye: [P2] Keep one authoritative import-conversion path

The imports have already been deserialized directly into `Vec<Import>` above, but these lines convert the same values through `MoonPkgJSON` and `pkg_json_imports_to_imports` again; the following loop then overwrites that result. This leaves two normalization paths and a separate string-key table governing which one wins. A future normalization change in `pkg_json_imports_to_imports` will no longer apply to ordinary DSL import blocks.

This conflicts with the repository's rule to keep one authoritative owner for each fact and remove redundant intermediate layers. Decode the DSL and legacy JSON representations separately where necessary, then normalize imports once and feed them into package construction without constructing and discarding a second import model.

crates/moon/src/cli/run.rs
```diff
@@ -592,6 +605,7 @@ fn build_package_executable(
     )
 }
 
+#[cfg(test)]
```
peter-jerry-ye: [P2] Have planner tests use the production selection phase

Making this entry point test-only leaves its path/package/backend selection as a separate copy of the production logic newly added in `build_package_executable`. Both implement the explicit-backend / selected-module-preference / default policy, but preferred-backend planner tests now exercise this copy rather than the production one. `PlanningFixture` also resolves every backend upfront, bypassing the new requirement to choose the backend before resolving its dependencies.

Please share the production selection phase between the command and the relevant tests, carrying its selected package and backend into resolution and planning. This preserves the existing narrow test boundary while avoiding duplicated policy that can drift without those tests failing.

crates/moon/src/rr_build/mod.rs
```diff
@@ -71,23 +71,53 @@ mod prebuild;
...
 
-/// Synchronize dependencies and return resolved project data.
+/// Synchronize dependencies and read module and package declarations.
 /// This step does not acquire the target-directory lock.
 pub(crate) fn sync_and_resolve_project(
     resolve_config: &ResolveConfig,
     dirs: &PackageDirs,
     user_log: &UserLog,
-) -> anyhow::Result<ResolveOutput> {
+) -> anyhow::Result<ProjectDeclarations> {
```
peter-jerry-ye, line 80: [P3] Name the new discovery phase explicitly

`sync_and_resolve_project` now returns `ProjectDeclarations`, with package relationships still unresolved. Several callers consequently use `resolve_output` and `*_from_resolved` names for declarations before solving them. Since this PR makes that phase boundary important, please rename the discovery entry point and its immediate callers/variables to make the unresolved state explicit. This is a non-blocking readability concern.

peter-jerry-ye (changes requested):
The declaration/resolution split and backend-specific dependency graphs are justified. Requesting changes for three P2 issues: duplicate active imports must be rejected before graph insertion, import conversion needs one authoritative normalization path, and planner tests must exercise the production backend-selection policy. There is also a non-blocking naming cleanup noted inline.

The import rule should be explicit: multiple blocks may be active, but each imported package may appear only once per backend within each import kind, regardless of alias. Disjoint conditions and different import kinds remain valid.

PR #2221, opened 2026-09-17
refactor(moonrun): own process input handles

peter-jerry-ye:
Process argument arrays, legacy environment blocks, and environment builders each coordinate a payload table with separately managed Host Keys. Creation and ownership transfer must keep both stores consistent.

Use the existing `Handles<T>` table for these three families on Unix and Windows. Each table owns registration, validation, and retirement together with its values, removing eight key-only helpers and the paired key/payload updates. Process operations retain their ownership rules: Unix spawn validates both inputs before consuming either, and environment-block transfer consumes its temporary source even when destination capacity is insufficient. Tests exercise successful transfer after rejected requests, wrong-family handles, stale handles, and invalid values including zero, null keys, and the reserved invalid Resource Handle.

The developer guide describes table mechanics and operation ownership. ADR 0009 records the shared namespace decision and its rationale. This continues #2219 with one domain-sized adoption of the existing table interface.


peter-jerry-ye:
@codex Review

PR #2235, opened 2026-09-18
fix(mooncake): prevent registry index pulls from hanging on large Git output

gaato:
- Related issues: None
- PR kind: Bugfix

## Summary

`git pull` can hang during registry index updates when its output fills the pipe buffer, for example after a large fast-forward. Use `wait_with_output()` to drain stdout and stderr while waiting, and include captured output when the command fails. Apply the same waiting pattern to `is_in_git_repo()`.

Add a regression test that pulls 2,000 new files from a local registry. The regression test and targeted registry-index and Git utility tests passed.

## Metadata


gaato:
I just signed the CCAA.

peter-jerry-ye (approved):
Thank you very much for the fix. Seems that we misused the API. I'll also try to replace the `git pull` later to avoid all the unnecessary work and output.

PR #2236, opened 2026-09-20
fix(manifest): remove unsupported subpackage declarations

peter-jerry-ye:
- Related issues: None
- PR kind: Bugfix

## Summary

Package manifests still expose subpackage declarations even though the current build engine does not implement their source-subset selection and linking semantics. Remove the package-level declaration and import-level flag from `MoonPkgJSON` and its import types, along with the generated schemas.

Preserve the existing handling of unknown package options and import-object fields. Removed `sub-package` and `sub_package` fields are ignored and cannot create subpackage targets. Regression coverage checks this behavior in JSON and `moon.pkg` options, including true and false flags across all three import lists and unrelated unknown fields.

Keep the internal `SubPackage` target and build machinery available for future implementation. The change is limited to the manifest boundary, its constructors, schemas, tests, and documentation.

## Metadata


peter-jerry-ye:
@codex review

